PR #1, opened 2019-05-22
fix: fix prettier on windows for trailing comma and cr lf

bertyhell:
Windows git checks out files as CRLF line-endings, and commits with LF line endings
The prettier complains about CRLF on window

Isn't the default airbnb styles to have trailing commas for objects and arrays?


.prettierrc
```diff
@@ -1,2 +1,5 @@
 printWidth: 100
 singleQuote: true
+endOfLine: auto
```
e2o: endOfLine: auto is de default, die optie mag je dus weglaten

PR #5, opened 2019-05-31
feat: add sort option to search results + clean filter of empty values

bertyhell:


bertyhell:
merged together with:
https://github.com/viaacode/avo2-client/pull/6

PR #21, opened 2019-07-03
Revert "feat: add collection detail page + rename detail to item"

e2o:
Reverts viaacode/avo2-client#20


e2o:
Mocht die nog niet mergen, excuses

PR #44, opened 2019-08-01
feat: added <Footer />

e2o:


src/shared/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,30 @@
+import React, { FunctionComponent } from 'react';
+import { Link } from 'react-router-dom';
+
+export const Footer: FunctionComponent = () => {
+	return (
+		<footer className="c-global-footer">
+			<div className="c-global-footer__inner o-container">
```
bertyhell: use Container component with className attribute:
https://github.com/viaacode/avo2-components/blob/master/src/components/Container/Container.tsx#L10
```
<Container mode="horizontal" className="c-global-footer__inner">
```

src/shared/components/Footer/Footer.tsx
```diff
@@ -0,0 +1,30 @@
...
+				<ul>
+					<li>
+						Een realisatie van <a href="https://viaa.be">VIAA.be</a>
+					</li>
+					<li>
+						<Link to="/terms-and-conditions">Gebruiksvoorwaarden</Link>
```
bertyhell: links should be in dutch: "/gebruiksvoorwaarden"

PR #46, opened 2019-08-01
fix: fix @viaa/types .git folder error by updating to @viaa/types@1.4.2

bertyhell:
This should allow you to do npm i without the need to delete the old typings first


andrych:
Ik heb nog de package-lock update gecommit

PR #47, opened 2019-08-02
fix: switch to apollo react client for getting graphql data

bertyhell:
To work efficiently with graphql, we will forward queries from the client via the proxy to graphql. The permissions will then also be configured by Bart in graphql. This saves a lot of complexity in the proxy and in the client since we won't have to create separate routes for each object we want to query.

This also allows us to use the integrated data store of the apollo graphql client, so we don't need to maintain reducers, selectors and actions for any data requests.

This PR requires this PR to be merged first: https://github.com/viaacode/avo2-proxy/pull/16


src/shared/components/DataComponent/DataQueryComponent.tsx
```diff
@@ -0,0 +1,47 @@
...
+	notFoundMessage?: string;
+}
+
+export const DataQueryComponent: FunctionComponent<DataQueryComponentProps> = ({
+	query,
+	variables,
+	resultPath,
+	renderData,
+	notFoundMessage = 'Het opgevraagde object werd niet gevonden',
+}: DataQueryComponentProps) => {
```
andrych: Deze mag weg

PR #96, opened 2019-09-12
AVO2-628: Adapt token retrieval to service + partial skinning.

benjaminnaesen:
I'll continue working on the skinning (for example that fullscreen button) on other branch. This one is getting too large.

- Gets player token from service instead of with redux.
- First show thumbnail, only load player when clicked.
- Partial skinning, basic color setup and shape.
- Expanded FlowPlayer component.


src/shared/components/FlowPlayer/FlowPlayer.scss
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	.is-fullscreen .fp-footer {
+		display: inline-block;
+	}
+
+	.fp-footer {
+		display: none;
+		max-width: calc(100% - 3.2rem);
+		top: 1em;
```
andrych: Waarom `em` gebruiken?

benjaminnaesen: Mono-style copy paste. Heb het gefixt het in de volgende skinning PR. :)

src/item/views/Item.tsx
```diff
@@ -1,4 +1,4 @@
-import { debounce } from 'lodash-es';
+import { debounce, get } from 'lodash-es';
```
andrych: Die get mag hier weg, er zijn nog wat ongebruikte imports

src/shared/components/FlowPlayer/FlowPlayer.tsx
```diff
@@ -1,18 +1,39 @@
-import React, { FunctionComponent, useEffect, useRef } from 'react';
+import React, { FunctionComponent, useEffect, useRef, useState } from 'react';
```
andrych: useState wordt nergens gebruikt

src/shared/components/FlowPlayer/FlowPlayer.tsx
```diff
@@ -21,7 +42,24 @@ export const FlowPlayer: FunctionComponent<FlowPlayerProps> = props => {
...
+	return src && poster ? (
+		<div
+			className="c-video-player"
+			data-player-id={(window as CustomWindow)._ENV_.FLOW_PLAYER_ID}
+			ref={videoPlayerRef}
+		/>
+	) : (
+		<div className="c-video-player">
+			<div className="c-video-player__item">
+				<img src={poster} />
```
andrych: Nog een alt attribute toevoegen

src/shared/components/FlowPlayer/FlowPlayer.tsx
```diff
@@ -1,18 +1,39 @@
...
+	title: string;
+	start?: number;
+	end?: number;
+	onInit?: () => void;
 }
 
-export const FlowPlayer: FunctionComponent<FlowPlayerProps> = props => {
+export const FlowPlayer: FunctionComponent<FlowPlayerProps> = ({ src, poster, title, onInit }) => {
 	const videoPlayerRef = useRef(null);
 	let player: any = null;
```
andrych: Linter warning: `Assignments to the 'player' variable from inside React Hook useEffect will be lost after each render. To preserve the value over time, store it in a useRef Hook and keep the mutable value in the '.current' property. Otherwise, you can move this variable directly inside useEffect`

PR #132, opened 2019-09-30
fix(login): avoid login loop when check-login call to api fails

bertyhell:
image


src/authentication/views/Login.tsx
```diff
@@ -28,11 +29,22 @@ const Login: FunctionComponent<LoginProps> = ({
 	getLoginState,
 }) => {
 	useEffect(() => {
-		if (!loginState && !loginStateLoading) {
+		console.log('login state error', loginStateError);
```
benjaminnaesen: Kill the log!

PR #134, opened 2019-10-01
fix(navigation): fix active navigation item color

bertyhell:
current situation:
wrong color for active state:
image

when going to assignments, the workspace navigation item isn't active:
image

After this PR both issues are resolved:
image

closes: https://district01.atlassian.net/browse/AVO2-635

### DIFF
--- a/src/shared/components/Navigation/Navigation.scss
+++ b/src/shared/components/Navigation/Navigation.scss
@@ -1,0 +1,61 @@
+.c-nav ul {
+	margin: 0;
+	padding: 0;
+}
+
+.c-nav li {
+	display: inline-block;
+	margin-right: 0.4rem;
+}
+
+.c-nav__item {
+	display: block;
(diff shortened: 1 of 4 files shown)


src/shared/components/Navigation/Navigation.scss
```diff
@@ -0,0 +1,61 @@
...
+	margin: 0;
+	padding: 0;
+}
+
+.c-nav li {
+	display: inline-block;
+	margin-right: 0.4rem;
+}
+
+.c-nav__item {
```
benjaminnaesen: Make it more sass-y.

```
.c-nav {
	&__item {
		&:hover, &:focus {

		}

		&--i {

		}
	}
}
```

bertyhell: Dit zal wss overschreven worden door andry zijn aanpassingen

PR #140, opened 2019-10-02
feat(assignments): add sorting, filtering and pagination

bertyhell:
image

closes https://district01.atlassian.net/browse/AVO2-421

de branch heeft per ongeluk de verkeerde jira issue reference


src/assignment/views/Assignments.tsx
```diff
@@ -38,8 +38,44 @@ const Assignments: FunctionComponent<AssignmentsProps> = ({ history }) => {
...
+				{ title: { _like: `%${uppercaseFilter}%` } },
+				{
+					assignment_assignment_tags: {
+						assignment_tag: { label: { _like: `%${uppercaseFilter}%` } },
+					},
+				},
+				{ class_room: { _like: `%${uppercaseFilter}%` } },
+				{ assignment_type: { _like: `%${uppercaseFilter}%` } },
+			];
+		} else {
```
benjaminnaesen: `else`-clause is unnecessary. `return` can just be placed after the if-statement.

bertyhell:
> Columns change size during sorting, is not directly related to this PR, but shouldn't happen. :)

Heb je een suggestie voor hoe we dit oplossen?

PR #163, opened 2019-10-14
fix(assignment-edit): Fix issue with assignment-edit eternal spinner

bertyhell:
issue was caused by multiple useEffects which all set the the currentAssignment state without waiting on eachother.

You could reproduce the issue as follows:
* search for a video
* make assignment

expected result: assignment edit screen
broken result: spinner kept spinning

### DIFF
--- a/src/assignment/views/AssignmentEdit.tsx
+++ b/src/assignment/views/AssignmentEdit.tsx
@@ -1,7 +1,7 @@
 import { useMutation } from '@apollo/react-hooks';
 import { ApolloQueryResult } from 'apollo-boost';
 import { DocumentNode } from 'graphql';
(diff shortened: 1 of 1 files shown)


src/assignment/views/AssignmentEdit.tsx
```diff
@@ -134,135 +137,177 @@ const AssignmentEdit: FunctionComponent<AssignmentEditProps> = ({
 	 *  Get query string variables and store them into the assignment state object
 	 */
 	useEffect(() => {
-		// Determine if this is an edit or create page
-		if (location.pathname.includes(RouteParts.Create)) {
-			setPageType('create');
+		initAssignmentData().then(() => {});
```
benjaminnaesen: `then` is unnecessary.

PR #164, opened 2019-10-15
fix(apollo-cache): clear cache after graphql mutations

bertyhell:
This solves the issue where you see old data in the client after you did a mutation

Steps to see the issue in the old version:
* search for a video
* click: add to collection
* create a new collection
* add
* search for another video
* add to collection

current result: dropdown list does not contain previously created collection
expected result: dropdown list contains all collections, even the one that was created just before this action

### DIFF
--- a/src/shared/services/data-service.ts
+++ b/src/shared/services/data-service.ts
@@ -13,7 +13,7 @@ export class ApolloCacheManager {
 	 * @param cache
 	 */
 	public static clearCollectionCache(cache: { [key: string]: any }) {
-		ApolloCacheManager.deleteCacheFromCache(cache, 'app_collection');
+		ApolloCacheManager.deleteFromCache(cache, 'app_collection');
 	}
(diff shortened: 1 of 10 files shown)


src/shared/services/data-service.ts
```diff
@@ -5,3 +5,28 @@ export const dataService = new ApolloClient({
...
+
+	/**
+	 * Clears all assignment related items from the cache
+	 * eg: app_assignment
+	 * @param cache
+	 */
+	public static clearAssignmentCache = (cache: { [key: string]: any }) =>
+		ApolloCacheManager.deleteCacheFromCache(cache, 'app_assignment');
+
+	private static deleteCacheFromCache(cache: { [key: string]: any }, keyPrefix: string) {
```
benjaminnaesen: I'd name this `deleteCache`, otherwise it's a bit confusing.

PR #173, opened 2019-10-17
Feature: event logging full pipeline

bertyhell:
this partly closes: https://district01.atlassian.net/browse/AVO2-647

Still need to trigger events for all existing functionality
but the server and client services are done, so you can report log events after this PR

You can see the log events in this graphql:
http://sc-events-qas-shared-components.apps.do-prd-okp-m0.do.viaa.be/console/data/schema/public/tables/events/browse


src/shared/helpers/formatters/avatar.tsx
```diff
@@ -5,13 +5,31 @@ import { Avatar, AvatarList } from '@viaa/avo2-components';
...
+	if ((obj as Avo.User.Profile).user) {
+		return obj;
+	}
+	return {
+		...((obj as Avo.User.User).profile || {}),
+		user: obj as Avo.User.User,
+	};
+};
+
+export const getInitialChar = (value: string | undefined | null | undefined) =>
```
andrych: `undefined` staat hier 2x

PR #182, opened 2019-10-28
fix(fragment-add-collection-modal): fix issue when adding fragment twice

bertyhell:
found a new bug while trying to reproduce: https://district01.atlassian.net/browse/AVO2-736

Steps:
add an item to a new collection
try adding the same item to a collection again

expected: the add to collection modal opens up, so you can choose between new or existing collection
current behavior: after adding the item to a collection the first time, the modal doesn't want to open anymore.

This issue has been fixed in this PR
The issue linked from jira was not reproducible

### DIFF
--- a/src/item/components/modals/FragmentAddToCollection.tsx
+++ b/src/item/components/modals/FragmentAddToCollection.tsx
@@ -100,7 +100,7 @@ const FragmentAddToCollection: FunctionComponent<FragmentAddToCollectionProps> =
 					toastService('Het ophalen van de bestaande collecties is mislukt', TOAST_TYPE.DANGER);
 				});
 		}
-	}, [isOpen]);
+	}, [isOpen, itemMetaData.duration]);
 
 	const setSelectedCollectionIdAndGetCollectionInfo = async (id: string) => {
 		try {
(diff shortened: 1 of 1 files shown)


src/item/components/modals/FragmentAddToCollection.tsx
```diff
@@ -150,10 +177,12 @@ const FragmentAddToCollection: FunctionComponent<FragmentAddToCollectionProps> =
 	const addItemToNewCollection = async () => {
 		// Disable "Toepassen" button
 		setIsProcessing(true);
+		console.log('adding to new collection');
```
benjaminnaesen: `console.log`

src/item/components/modals/FragmentAddToCollection.tsx
```diff
@@ -72,9 +71,37 @@ const FragmentAddToCollection: FunctionComponent<FragmentAddToCollectionProps> =
...
 
+	useEffect(() => {
+		if (isOpen) {
+			// Reset the state
+			setCreateNewCollection(false);
+			setSelectedCollectionId('');
+			setSelectedCollection(undefined);
+			setNewCollectionTitle('');
+			setFragmentStartTime(0);
+			setFragmentEndTime(toSeconds(itemMetaData.duration) || 0);
```
andrych: `itemMetaData.duration` moet nog mee in de dependency array

PR #185, opened 2019-10-29
fix(cut-fragment-modal): start_oc never gets saved

bertyhell:
### Steps to reproduce the bug before this PR
* go to a collection
* edit a fragment start time
* save the collection
* reload the page
* open the cut fragment modal

### Expected result:
the previously set start_oc cut time should be shown

### Current result:
the start_oc _cut time is set to 0

### Cause
This is caused by the asynchronous nature of setState functions. So try to avoid this pattern in the future:
Calling 2 functions that set the state, right after eachother, where the functions internally take a copy of the current state, and then set that state at the end of the function.
https://github.com/viaacode/avo2-client/blob/3c96c4a7dcc0098d6e6e78188620852429858379/src/collection/components/modals/CutFragmentModal.tsx#L47-L48

What happened was:
* update start_oc function is called
* it clones the currentCollection, modifies it and calls setState with the copy
* update start_end function is called
* it clones the currentCollection, modifies it and calls setState with the copy
* the first setState completes and sets the state with the modified start_oc
* the second setState completes and sets the state with the modified end_oc (start_oc is overridden here)

### DIFF
--- a/src/collection/components/modals/CutFragmentModal.tsx
+++ b/src/collection/components/modals/CutFragmentModal.tsx
@@ -22,11 +22,17 @@ import { fetchPlayerTicket } from '../../../shared/services/player-ticket-servic
 import toastService, { TOAST_TYPE } from '../../../shared/services/toast-service';
 import { getValidationErrorsForStartAndEndTime } from '../../helpers/validation';
 
+export interface FragmentPropertyUpdateInfo {
+	value: string | number | boolean | null;
+	fieldName: keyof Avo.Collection.Fragment;
+	fragmentId: number;
+}
+
 interface CutFragmentModalProps {
 	isOpen: boolean;
 	onClose: () => void;
 	itemMetaData: Avo.Item.Item;
-	updateFragmentProperty: (value: any, fieldName: string, fragmentId: number) => void;
+	updateFragmentProperties: (updateInfos: FragmentPropertyUpdateInfo[]) => void;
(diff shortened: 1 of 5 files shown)


src/collection/components/modals/CutFragmentModal.tsx
```diff
@@ -89,8 +95,10 @@ const CutFragmentModal: FunctionComponent<CutFragmentModalProps> = ({
 		const start = toSeconds(fragmentStartTimeString, true);
 		const end = toSeconds(fragmentEndTimeString, true);
 
-		updateFragmentProperty(start, 'start_oc', fragment.id);
-		updateFragmentProperty(end, 'end_oc', fragment.id);
+		updateFragmentProperties([
+			{ value: start, fieldName: 'start_oc' as const, fragmentId: fragment.id },
```
andrych: Waarom hier die `as const`? (Heb dat nog niet eerder gezien 😄)

bertyhell: omdat typescript anders denkt dat dat een generic string kan zijn. Door te zeggen dat het een const is, kan hij dat type ('start_oc') resolven met het type: (keyof Avo.Collection.Fragment)

Toegevoegd aan typescript 3.4: https://devblogs.microsoft.com/typescript/announcing-typescript-3-4-rc/
Onder de titel: const assertions

PR #192, opened 2019-10-30
fix(cut-fragment): not update start/end when no cut.

benjaminnaesen:
If we always apply the start/end time, we will never know whether a fragment was cut. Hence don't always apply the start/end. Only if cut.

### DIFF
--- a/src/item/components/modals/FragmentAddToCollection.tsx
+++ b/src/item/components/modals/FragmentAddToCollection.tsx
@@ -127,6 +127,10 @@ const FragmentAddToCollection: FunctionComponent<FragmentAddToCollectionProps> =
 		setIsProcessing(true);
 
 		try {
+			const hasCut = !(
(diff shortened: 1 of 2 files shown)


src/item/components/modals/FragmentAddToCollection.tsx
```diff
@@ -127,17 +127,21 @@ const FragmentAddToCollection: FunctionComponent<FragmentAddToCollectionProps> =
 		setIsProcessing(true);
 
 		try {
+			const hasCut = !(
+				fragmentEndTime === toSeconds(itemMetaData.duration) && fragmentStartTime === 0
```
bertyhell: dit lijkt me logischer
```fragmentEndTime !== toSeconds(itemMetaData.duration) || fragmentStartTime !== 0```

benjaminnaesen: Achteraf gezien, mij ook ;)

PR #261, opened 2019-12-04
fix(routing): switch collection to collections route + cleanup

bertyhell:
This is needed to bring the route in line with avo1

### DIFF
--- a/src/shared/components/DateRangeDropdown/DateRangeDropdown.test.tsx
+++ b/src/shared/components/DateRangeDropdown/DateRangeDropdown.test.tsx
@@ -2,6 +2,7 @@ import { mount, shallow } from 'enzyme';
 import React from 'react';
 
 import { action } from '@storybook/addon-actions';
+import DateRangeDropdown from './DateRangeDropdown';
 
 describe('<Checkbox />', () => {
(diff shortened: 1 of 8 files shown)


src/shared/components/DateRangeDropdown/DateRangeDropdown.test.tsx
```diff
@@ -2,6 +2,7 @@ import { mount, shallow } from 'enzyme';
 import React from 'react';
 
 import { action } from '@storybook/addon-actions';
+import DateRangeDropdown from './DateRangeDropdown';
```
benjaminnaesen:
```
import React from 'react';
import { action } from '@storybook/addon-actions';

import DateRangeDropdown from './DateRangeDropdown';
```

src/shared/constants/index.ts
```diff
@@ -1,2 +1,2 @@
 export { LANGUAGES } from './languages';
-export { CONTENT_TYPE_TO_ROUTE, ROUTE_PARTS } from './routes';
+export { ROUTE_PARTS } from '../constants/routes';
```
andrych: `... from './routes';`

PR #276, opened 2019-12-11
AVO2-828: added RichTextTwoColumns cb config

benjaminnaesen:


src/admin/content-block/helpers/generators/defaults.ts
```diff
@@ -14,27 +16,43 @@ export const FORM_STATE_DEFAULTS = (
...
+	label,
+	editorType: ContentBlockEditor.AlignSelect,
+	editorProps: {
+		options: ALIGN_OPTIONS,
+	},
+});
+
+export const TEXT_FIELD = (
+	label: string = 'Tekst',
+	emptyFieldValidatorMessage: string = 'Tekstveld moet verplicht ingevuld zijn.'
```
bertyhell: type string is overbodig

https://stackoverflow.com/questions/46799828/tslint-type-trivially-inferred-why-is-it-bad-practice-to-include-the-type-he

src/admin/content-block/helpers/generators/rich-text-two-columns.ts
```diff
@@ -0,0 +1,29 @@
...
+	secondColumnContent: '',
+	...FORM_STATE_DEFAULTS(ContentBlockBackgroundColor.White, ContentBlockType.Heading),
+});
+
+export const RICH_TEXT_TWO_COLUMNS_BLOCK_CONFIG = (): ContentBlockConfig => ({
+	name: 'Tekst (2 kolommen)',
+	initialState: INITIAL_RICH_TEXT_TWO_COLUMNS_BLOCK_STATE,
+	fields: {
+		firstColumnContent: TEXT_FIELD(
+			'Kolom 1 Tekst',
```
bertyhell: "Kolom 1" is reeds duidelijk genoeg

PR #280, opened 2019-12-16
fix(login-flow): rework login flow

bertyhell:
rework login flow:

image

image

image


src/authentication/components/LoginOptions.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+export interface LoginOptionsProps extends RouteComponentProps {
+	onOptionClicked?: () => void;
+}
+
+const LoginOptions: FunctionComponent<LoginOptionsProps> = ({
+	history,
+	location,
+	onOptionClicked = () => {},
+}) => {
```
benjaminnaesen: Je kunt hier direct returnen.

PR #285, opened 2019-12-17
fix(profile): fix saving organizations

bertyhell:
This PR requires proxy changes: https://github.com/viaacode/avo2-proxy/pull/70

Fixing this now since i couldn't test this before because of the select onchange bug


src/settings/settings.service.ts
```diff
@@ -1,4 +1,5 @@
 import { Avo } from '@viaa/avo2-types';
+import { ClientEducationOrganization } from '../shared/services/education-organizations-service';
```
andrych: Hier nog een enter tussen

PR #286, opened 2019-12-18
feat(smartschool-login): allow login/logout for teacher with smartschool

bertyhell:
closes: https://district01.atlassian.net/browse/AVO2-158

This PR has a proxy twin: https://github.com/viaacode/avo2-proxy/pull/71

After the user links a smartschool account to their profile in the profile settings, they should be able to login/logout using smartschool
image


benjaminnaesen:
After logging in, user gets redirected to the search page. Is that supposed to be? Feels odd UX-wise.

bertyhell:
> After logging in, user gets redirected to the search page. Is that supposed to be? Feels odd UX-wise.
Ja, dat was by design. Maar mss is het beter als we dit aanpassen naar de home page.

PR #311, opened 2020-01-16
Feature: create nav bars using navigation items in the database (2)

bertyhell:


bertyhell:
together with https://github.com/viaacode/avo2-client/pull/309

PR #318, opened 2020-01-20
AVO2-830: Content Picker (v2), added collections type.

benjaminnaesen:
To test this, you need to add "collection" to the selectableTypes array in MenuEditForm. 

Saving with a collection won't work yet as we need to discuss the url path/links first.

Heb ook "het optimaliseren van de ophaallimit" nog op mijn todo maar dat is nu minder dringend. Dat doe ik als ik tussenin tijd vind.


src/admin/content/components/ContentPicker/ContentPicker.tsx
```diff
@@ -38,7 +38,7 @@ export const ContentPicker: FunctionComponent<ContentPickerProps> = ({
 		if (currentTypes && currentTypes.length) {
 			setLoading(true);
 
-			const maxPerType = 20 / currentTypes.length;
+			const maxPerType = Math.floor(20 / currentTypes.length);
```
benjaminnaesen: I'll invest time in a better way to do this limits later. Not prio.

benjaminnaesen:
Replaced by https://github.com/viaacode/avo2-client/pull/320

PR #326, opened 2020-01-21
fix(content-page): load new content page if route changes

bertyhell:
closes: https://district01.atlassian.net/browse/AVO2-952

### DIFF
--- a/src/content-page/views/ContentPage.tsx
+++ b/src/content-page/views/ContentPage.tsx
@@ -1,4 +1,4 @@
-import React, { FunctionComponent, useEffect, useState } from 'react';
+import React, { FunctionComponent, useCallback, useEffect, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 import { withRouter } from 'react-router';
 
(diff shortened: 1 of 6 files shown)


src/content-page/views/ContentPage.tsx
```diff
@@ -18,8 +18,16 @@ interface ContentPageDetailProps extends DefaultSecureRouteProps {}
 const ContentPage: FunctionComponent<ContentPageDetailProps> = ({ match, user }) => {
 	const [t] = useTranslation();
 
+	const getCurrentPath = () => `/${(match.params as any).path}`;
```
andrych: Hierboven `interface ContentPageDetailProps extends DefaultSecureRouteProps<{ path: string }> {}` dan mag die `as any` weg.

andrych: Also: `const getCurrentPath = useCallback(() => `/${match.params.path}`, [match.params.path]);` om onderstaande warning op te lossen.

src/content-page/views/ContentPage.tsx
```diff
@@ -18,8 +18,16 @@ interface ContentPageDetailProps extends DefaultSecureRouteProps {}
...
+
 	// State
-	const [path] = useState<string>(`/${(match.params as any)['path']}`);
+	const [path, setPath] = useState<string>(getCurrentPath());
+
+	useEffect(() => {
+		if (path !== getCurrentPath()) {
+			setPath(getCurrentPath());
+		}
+	}, [match]);
```
andrych: Warning: React Hook useEffect has missing dependencies: 'getCurrentPath' and 'path'. `match` mag weg.

src/content-page/views/ContentPage.tsx
bertyhell: ok, overal gefixed

PR #352, opened 2020-01-29
feat(bundle-detail): show published bundles that contain collection

bertyhell:
image


src/collection/views/CollectionDetail.tsx
```diff
@@ -72,6 +72,7 @@ import { FragmentList, ShareCollectionModal } from '../components';
 import AddToBundleModal from '../components/modals/AddToBundleModal';
 
 import './CollectionDetail.scss';
+import { Link } from 'react-router-dom';
```
benjaminnaesen: Import order

PR #356, opened 2020-01-30
Fix issue with use_custom_fields in collection edit page

bertyhell:
steps to reproduce issue before this PR:

edit collection
enable custom fields switch
add some text
save collection
reload page
expected: custom fields switch is still enabled

current: custom fields switch is back in the off position


src/collection/components/CollectionOrBundleEdit.tsx
```diff
@@ -241,7 +240,8 @@ const CollectionOrBundleEdit: FunctionComponent<CollectionOrBundleEditProps> = (
 	};
 
 	// Update individual property of fragment
-	const updateFragmentProperties = (updateInfos: FragmentPropertyUpdateInfo[]) => {
+	const onFragmentChanged = (fragment: Avo.Collection.Fragment) => {
+		console.log('fragment props updated: ', fragment);
```
andrych: Deze mag weg

PR #372, opened 2020-02-04
fix(error-view): allow not logged in user to see error view from server

bertyhell:
The content block route was before the error view route, so when the server redirects to the error view page, the content page route would be activated, but the user needs to be logged in for that, so the user gets redirected to the login screen

Also added one extra action button: helpdesk

image


src/error/views/ErrorView.tsx
```diff
@@ -1,20 +1,31 @@
...
 import queryString from 'querystring';
 import React, { FunctionComponent, ReactNode } from 'react';
+import { useTranslation } from 'react-i18next';
 import { match, RouteComponentProps, withRouter } from 'react-router';
 
-import { Blankslate, Button, Container, IconName } from '@viaa/avo2-components';
+import {
+	Blankslate,
+	Button,
+	ButtonGroup,
```
andrych: 'ButtonGroup' is defined but never used

src/error/views/ErrorView.tsx
```diff
@@ -45,18 +56,37 @@ const ErrorView: FunctionComponent<ErrorViewProps> = ({
...
+								{buttons.includes('home') && (
+									<Button
+										onClick={() => redirectToClientPage(APP_PATH.LOGGED_IN_HOME, history)}
+										label={t('error/views/error-view___ga-terug-naar-de-homepagina')}
+									/>
+								)}
+								{buttons.includes('helpdesk') && (
+									<Button
+										type="danger"
+										onClick={() => (window as any).zE('webWidget', 'toggle')}
```
andrych: Je kan die toevoegen aan de Window interface in `shared/types/global.d.ts`, dan moet die as any niet.

PR #381, opened 2020-02-10
fix(item-detail): fix infinite spinner after adding item to collection

bertyhell:
Steps:
* go to an item detail page
* add it to an existing collection

expected result:
the add to collection modal closes and the item detail is visible again

current result:
The add to collection modal closes and you see an infinite spinner

### DIFF
--- a/src/item/views/Item.tsx
+++ b/src/item/views/Item.tsx
@@ -158,8 +158,8 @@ const Item: FunctionComponent<ItemProps> = ({ history, match, location, user, ..
 			} catch (err) {
 				console.error(
 					new CustomError('Failed to check permissions or get item from graphql', err, {
-						itemId: match.params.id,
 						user,
+						itemId: match.params.id,
 					})
 				);
 				setLoadingInfo({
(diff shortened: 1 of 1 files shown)


src/item/views/Item.tsx
```diff
@@ -72,12 +72,94 @@ const Item: FunctionComponent<ItemProps> = ({ history, match, location, user, ..
...
+			});
+		}
+	}, [item, setLoadingInfo]);
+
+	/**
+	 * Load item from database
+	 */
+	useEffect(() => {
+		checkPermissionsAndGetItem();
+	}, [itemId, setItem]);
```
andrych: React Hook useEffect has a missing dependency: 'checkPermissionsAndGetItem'.

src/item/views/Item.tsx
```diff
@@ -72,12 +72,94 @@ const Item: FunctionComponent<ItemProps> = ({ history, match, location, user, ..
 
 	const [t] = useTranslation();
 
-	const [itemId] = useState<string | undefined>(match.params.id);
+	const [itemId] = useState<string>(match.params.id);
```
andrych: Moet niet in state: `const { id: itemId } = match.params;`

PR #390, opened 2020-02-17
AVO2-992: Content Picker - rework + display existing values

benjaminnaesen:


src/admin/shared/components/ContentPicker/ContentPicker.tsx
```diff
@@ -4,67 +4,62 @@ import { useTranslation } from 'react-i18next';
...
 	errors?: string | string[];
+	currentSelection?: PickerItem;
 }
 
 const ContentPicker: FunctionComponent<ContentPickerProps> = ({
-	initialControls = 'content',
 	selectableTypes,
 	onSelect,
 	errors = [],
+	currentSelection,
```
bertyhell: default to "CONTENT" type

src/admin/content/content.const.ts
```diff
@@ -90,39 +82,6 @@ export const CONTENT_DETAIL_TABS: TabProps[] = [
...
-		value: 'BUNDLE',
-		label: i18n.t('admin/content/content___bundels'),
-		disabled: false,
-		fetch: fetchBundles,
-	},
-];
-
 export const CONTENT_WIDTH_OPTIONS = [
 	{ label: 'Kies een content breedte', value: '', disabled: true },
 	{ label: 'Max. (1300px)', value: 'REGULAR' },
```
bertyhell: translatable

PR #402, opened 2020-02-18
Add button types

bertyhell:
I found this branch quite a while back into the git history, but it still seems to be relevant.
It adds all button types so the user can select any button style.

Currently you can only select primary and secondary styles


src/admin/content-block/content-block.types.ts
```diff
@@ -12,7 +12,17 @@ export type AlignOptions = 'left' | 'right' | 'center';
 
 export type HeadingLevelOptions = 'h2' | 'h3' | 'h4';
 
-export type ButtonTypeOptions = 'primary' | 'secondary';
+export type ButtonTypeOptions =
```
andrych: Deze best vanuit de components exporten, zet hier dan nog een TODO

bertyhell: done https://github.com/viaacode/avo2-components/pull/232

PR #441, opened 2020-03-05
fix(auth): show error message on check login error

bertyhell:
for instance when you're on the train and you can connect the first time, but then you lose connection

image


src/authentication/components/SecuredRoute.tsx
```diff
@@ -1,21 +1,25 @@
...
 import { Dispatch } from 'redux';
 
 import { Flex, Spacer, Spinner } from '@viaa/avo2-components';
 import { Avo } from '@viaa/avo2-types';
 
 import { APP_PATH } from '../../constants';
 import { Profile } from '../../settings/components';
+import { buildLink } from '../../shared/helpers';
 
+import { useTranslation } from 'react-i18next';
```
andrych: Import order

src/error/views/ErrorView.tsx
```diff
@@ -47,18 +48,16 @@ const ErrorView: FunctionComponent<ErrorViewProps> = ({
...
-		(location.search || '').substring(1)
-	) as ErrorViewQueryParams;
+	const queryParams = queryString.parse((location.search || '').substring(1));
 	const errorMessage: string =
-		queryParams.message ||
+		(queryParams.message as string) ||
 		message ||
 		i18n.t('error/views/error-view___de-pagina-werd-niet-gevonden');
-	const errorIcon: IconName = queryParams.icon || icon || 'search';
+	const errorIcon: IconName = (queryParams.icon || icon || 'search') as IconName;
```
andrych: `: IconName` mag weg dan

PR #447, opened 2020-03-06
Bugfix/content blocks perf

andrych:
**Problem**: When adding many content-blocks on the edit page, editing in the input, textarea and wysiwig field-editors was practically unusable (slow performance).

**Cause**: Each content-block form gets re-rendered when a single one gets updated. This was really visible when typing in one of the editors mentioned above.

**Solution**: 
* Add `React.memo` to the smallest parts (form-group and field-editor components) in the chain
* Add `React.memo` and custom compare function for the form component. This is where we got the biggest performance gain.
* Debounce TextInput and Textarea editors onChange handlers to trigger less dispatches

### DIFF
--- a/src/admin/content-block/components/ContentBlockForm/ContentBlockForm.tsx
+++ b/src/admin/content-block/components/ContentBlockForm/ContentBlockForm.tsx
@@ -268,7 +268,8 @@ const ContentBlockForm: FunctionComponent<ContentBlockFormProps> = ({
 
 export default React.memo(ContentBlockForm, (prevProps, nextProps) => {
 	let areEqual = false;
-
+	// We don't need to check the other props because they are functions
+	// which will never change
 	if (
 		isEqual(prevProps.config, nextProps.config) &&
 		prevProps.isAccordionOpen === nextProps.isAccordionOpen &&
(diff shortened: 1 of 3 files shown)


src/admin/content-block/components/ContentBlockForm/ContentBlockForm.tsx
```diff
@@ -266,4 +266,18 @@ const ContentBlockForm: FunctionComponent<ContentBlockFormProps> = ({
...
+	if (
+		isEqual(prevProps.config, nextProps.config) &&
+		prevProps.isAccordionOpen === nextProps.isAccordionOpen &&
+		prevProps.blockIndex === nextProps.blockIndex &&
+		prevProps.length === nextProps.length
+	) {
+		areEqual = true;
+	}
+	// The component will rerender when false is returned
+	return areEqual;
```
bertyhell:
```typescript
return isEqual(prevProps.config, nextProps.config) &&
		prevProps.isAccordionOpen === nextProps.isAccordionOpen &&
		prevProps.blockIndex === nextProps.blockIndex &&
		prevProps.length === nextProps.length;
```

andrych: Refactored

PR #448, opened 2020-03-09
Feature: Filter Table

bertyhell:
Table that handles all filtering, sorting, pagination and linking all those to the query params. All you have to do is react to onTableStateChanged

We should use this component for all data tables in the management dashboard.

This PR replaces and fixes all remarks of https://github.com/viaacode/avo2-client/pull/438

image


src/admin/shared/components/Sidebar/Sidebar.tsx
```diff
@@ -1,9 +1,10 @@
 import classnames from 'classnames';
-import React, { FunctionComponent } from 'react';
+import * as H from 'history';
+import React, { FunctionComponent, ReactElement } from 'react';
 import { Link, NavLink } from 'react-router-dom';
-
 import { NavigationItemInfo } from '../../../../shared/types';
 
+import { flatten } from 'lodash-es';
```
benjaminnaesen: Import order.

bertyhell: fixed in: https://github.com/viaacode/avo2-client/commit/705402c33cc42a0e630aa80488a44d3c4eba9439

PR #473, opened 2020-03-18
feat(share-email): close when done

benjaminnaesen:
- When item is being shared by email, close modal on success.


bertyhell:
fixed issue with send route: https://github.com/viaacode/avo2-proxy/pull/104

PR #475, opened 2020-03-18
chore(package): add peer dep. to keep track of compatible proxy version

bertyhell:
This way we'll be able to tell for every client version, which proxy version is compatible with it


package.json
```diff
@@ -105,6 +105,9 @@
 		"tslint-plugin-prettier": "^2.0.1",
 		"typescript": "3.5.3"
 	},
+  "peerDependencies": {
```
benjaminnaesen: indent

PR #507, opened 2020-03-27
fix(quality-label): remove duplicate fetch quality labels query

bertyhell:


src/collection/components/CollectionOrBundleEditAdmin.tsx
```diff
@@ -19,6 +19,7 @@ import { ToastService } from '../../shared/services';
 
 import { CollectionService } from '../collection.service';
 import { CollectionAction } from './CollectionOrBundleEdit';
+import { QualityLabel } from '../collection.types';
```
andrych: Import order

PR #532, opened 2020-04-02
feat(rich-text-block): add table button + fix main styles class mono

bertyhell:
closes: https://meemoo.atlassian.net/browse/AVO-91

missing border for last row is fixed in: https://github.com/viaacode/avo2-components/pull/273

image

image

image


src/admin/content-block/helpers/generators/defaults.ts
```diff
@@ -11,6 +11,8 @@ import {
 } from '../../../shared/types';
 
 import { GET_ALIGN_OPTIONS, GET_BACKGROUND_COLOR_OPTIONS } from '../../content-block.const';
+import { WYSIWYGProps } from '@viaa/avo2-components';
```
benjaminnaesen, line 14: Import order.

PR #593, opened 2020-04-23
fix(content-picker): fix target switch resetting value

benjaminnaesen:
- Switching the target toggle made it that the value was an object instead of the id of whatever was selected, hence the link didn't work anymore, and the second content picker select field showed empty as the object was not part of the options.


src/admin/shared/components/ContentPicker/ContentPicker.tsx
```diff
@@ -194,7 +194,7 @@ export const ContentPicker: FunctionComponent<ContentPickerProps> = ({
 		} else if (selectedType.picker === 'TEXT_INPUT') {
 			newValue = input;
 		} else if (selectedType.picker === 'SELECT' && selectedItem) {
-			newValue = (selectedItem as PickerItem).value;
+			newValue = get(selectedItem as PickerItem, 'value.value');
```
bertyhell, line 197: "as PickerItem" is not needed anymore

PR #635, opened 2020-05-04
feat(bundle): save description_long

benjaminnaesen:
closes: AVO-197

- Ability to save description_long for bundles.
- Max length of 1200 on description_long.
- Use description_long for bundle detail page.


bertyhell:
check max length at save:
image

PR #677, opened 2020-05-14
fix: subscript & superscript

benjaminnaesen:
Superscript and subscript didn't work in preview because mono styles reset their vertical-align. 🙈


src/styles/generic/_reset.scss
```diff
@@ -9,7 +9,7 @@ html, body, div, span, applet, object, iframe,
 h1, h2, h3, h4, h5, h6, p, blockquote, pre,
 a, abbr, acronym, address, big, cite, code,
 del, dfn, em, img, ins, kbd, q, s, samp,
-small, strike, strong, sub, sup, tt, var,
+small, strike, strong, tt, var,
```
bertyhell, line 12: mss fixen we dit dan best ook is de components?
https://github.com/viaacode/avo2-components/blob/master/src/styles/generic/_reset.scss#L12

PR #680, opened 2020-05-18
fix(AVO-798): map text area by value prop not values prop

bertyhell:
closes: https://meemoo.atlassian.net/browse/AVO-798

image


bertyhell:
> I'm not seeing a difference code-wise? Please explain. 😅

```
		case ContentBlockEditor.TextArea:
		case ContentBlockEditor.UserGroupSelect:
```
is veranderd in:
```
		case ContentBlockEditor.UserGroupSelect:
```

Dus de textarea control w nu op de default manier behandeld via onChange en de value prop.
Voordien door de switch falltrough statement was dit: onChange en **values**

PR #718, opened 2020-05-29
Use correct video still after cut

bertyhell:
closes: https://meemoo.atlassian.net/browse/AVO-129

requires proxy PR: https://github.com/viaacode/avo2-proxy/pull/146

image


bertyhell:
already in release/v0.36.0

PR #721, opened 2020-06-02
feat(search-block): add search block and replace static home by cb page

bertyhell:
closes: https://meemoo.atlassian.net/browse/AVO-857

image
image
image


bertyhell:
already in release/v0.36.0

PR #922, opened 2020-08-03
fix(translations): only update changed translations in the database

bertyhell:
closes: https://meemoo.atlassian.net/browse/AVO-1240

This fixes 90% of the overwrite issues with translations, by only saving the changed values to the database.
Users can still overwrite the same translation if the change it at the same time.

image

After step 4 this is the result:
image

After step 6 this is the result:
image

### DIFF
--- a/src/admin/translations/views/TranslationsOverview.tsx
+++ b/src/admin/translations/views/TranslationsOverview.tsx
@@ -1,5 +1,5 @@
-import { flatten, fromPairs, get, groupBy, map } from 'lodash-es';
-import React, { FunctionComponent, useEffect, useState } from 'react';
+import { flatten, fromPairs, get, groupBy, isNil, map } from 'lodash-es';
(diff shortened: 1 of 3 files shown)


src/admin/translations/views/TranslationsOverview.tsx
```diff
@@ -47,14 +51,43 @@ const TranslationsOverview: FunctionComponent<TranslationsOverviewProps> = () =>
 		// convert translations to db format and save translations
 		const promises: any = [];
 
-		convertDataToTranslations(translations).forEach((context: any) => {
+		const freshTranslations = convertTranslationsToData(await fetchTranslations());
+
+		const updatedTranslations = freshTranslations.map((freshTranslation: [string, string]): [
```
benjaminnaesen, line 56: Hier wordt 2 x `[string, string]` gebruikt. Ik zou dit herbruikbaar maken als `Translation` ofzo.

PR #936, opened 2020-08-06
feat(admin-item-detail): show organisation name instead of id

benjaminnaesen:
image

Resolves: https://meemoo.atlassian.net/browse/AVO-685


bertyhell:
best eerst deze PR erin mergen, dan kan alles ineens in 1.2.0 gemerged worden: https://github.com/viaacode/avo2-client/pull/938

PR #956, opened 2020-08-10
feat(coll-edit-detail): show table with assignments that contain coll.

bertyhell:
closes: https://meemoo.atlassian.net/browse/AVO-759

image

image

image


benjaminnaesen:
Wel nog merge conflct ^

PR #970, opened 2020-08-12
fix(content-edit): move onPaste event to body instead of contentEdit div

bertyhell:
closes: https://meemoo.atlassian.net/browse/AVO-936

image


src/admin/content/views/ContentEdit.tsx
```diff
@@ -145,6 +145,43 @@ const ContentEdit: FunctionComponent<ContentEditProps> = ({ history, match, user
 		}
 	}, [id, user, t]);
 
+	const onPasteContentBlock = useCallback(
+		(e: any) => {
```
benjaminnaesen, line 149: Not 100% sure, maar ik herinner me het type `ClipboardEvent`. Seems like I forgot to use it.

PR #1001, opened 2020-08-27
feat(item-detail): allow depublish item with reason or replacement item

bertyhell:
closes:
* https://meemoo.atlassian.net/browse/AVO-699
* https://meemoo.atlassian.net/browse/AVO-700

only the message on the detail page of a collection to indicate to the owner that a fragment was replaced by another and that the start en end time has been reset, has not been added yet. For that we wait on: https://meemoo.atlassian.net/browse/DEV-1166

add item to collection and cut it:
image

depublish item and replace it by another:
image

item is replaced in the collection and the cut is reset:
image

depublish with a reason shows this screen when you go directly to the item url:
image


benjaminnaesen (approved):

benjaminnaesen:
Deepcode ^

PR #1005, opened 2020-08-27
feat(filter-table): allow selection of columns, update admin sidebar

bertyhell:
part of: https://meemoo.atlassian.net/browse/AVO-1305

By linking the columns to the url, we can control the columns very precisely. For instance we can even choose the order of the columns. This url has the title as the second column:
http://localhost:8080/beheer/collecties?columns=author_user_group~title~is_public~collection_labels&page=0

image

image


benjaminnaesen (approved):

benjaminnaesen:
Conflict.

PR #1091, opened 2020-10-28
feat(admin-layout): make admin layout topbar same width as body

bertyhell:
makes admin layout header width match body and reduces the amount of code at the same time 💪 

|before|after|
|:---:|:---:|
|image|image|

|before|after|
|:---:|:---:|
|image|image|

|before|after|
|:---:|:---:|
|image|image|

### DIFF
--- a/src/admin/menu/views/MenuDetail.tsx
+++ b/src/admin/menu/views/MenuDetail.tsx
@@ -306,7 +306,7 @@ const MenuDetail: FunctionComponent<MenuDetailProps> = ({ history, match }) => {
 						</Flex>
 					</Spacer>
 					<DeleteObjectModal
-						deleteObjectCallback={() => handleDelete()}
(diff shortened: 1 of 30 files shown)


src/admin/menu/views/MenuDetail.tsx
```diff
@@ -233,99 +233,88 @@ const MenuDetail: FunctionComponent<MenuDetailProps> = ({ history, match }) => {
...
+									})
+								}
+								type="primary"
 							/>
-						</Container>
-					</Container>
+						</Flex>
+					</Spacer>
+					<DeleteObjectModal
+						deleteObjectCallback={() => handleDelete()}
```
benjaminnaesen, line 314: `deleteObjectCallback={handleDelete}`

benjaminnaesen (approved):

PR #1159, opened 2021-01-14
fix(assignment-detail): only create one response entry

bertyhell:
closes: https://meemoo.atlassian.net/browse/AVO-1371

fix issue where assignmentUuid isn't correctly passed to the proxy, causing the collection content of an assignment to not show up

make sure only one assignment_response entry is created by removing the caching for the request to check if a response already exists and fixing the issue where the user object is stored twice in the redux store.

### DIFF
--- a/src/authentication/store/actions.tsx
+++ b/src/authentication/store/actions.tsx
@@ -58,7 +58,7 @@ function checkIfSessionExpires(expiresAt: string) {
 	}
 }
 
-let loading: boolean = false;
+let loading = false;
 
 export const getLoginStateAction = () => {
 	return async (dispatch: Dispatch, getState: any): Promise<Action | null> => {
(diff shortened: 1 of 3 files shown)


src/authentication/store/actions.tsx
```diff
@@ -103,9 +112,9 @@ export const setLoginError = (): SetLoginErrorAction => ({
 	error: true,
 });
 
-export const setLoginLoading = (): SetLoginLoadingAction => ({
+export const setLoginLoading = (isLoading: boolean): SetLoginLoadingAction => ({
```
benjaminnaesen, line 115: Als je het hier `loading` noemt, kan je hieronder `loading: isLoading,` vervangen door `loading,`.

benjaminnaesen (approved):

src/authentication/store/actions.tsx
bertyhell: nee want er is al een globale variabele die loading heet. Dan zou het onduidelijk zijn of ik de functie parameter bedoel of the globale variabele

PR #1213, opened 2021-06-22
fix(marcom): delete entry

benjaminnaesen:
https://meemoo.atlassian.net/browse/AVO-1552

image

### DIFF
--- a/src/collection/collection.const.ts
+++ b/src/collection/collection.const.ts
@@ -65,7 +65,7 @@ export const GET_MARCOM_ENTRY_TABLE_COLUMNS: () => TableColumn[] = () => [
 		id: 'external_link',
 	},
 	{
-		label: i18n.t('Acties'),
+		label: '',
 		id: 'actions',
 	},
 ];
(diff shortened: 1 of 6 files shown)


src/collection/collection.service.ts
```diff
@@ -1294,6 +1295,27 @@ export class CollectionService {
...
+					id,
+				},
+				mutation: DELETE_MARCOM_ENTRY,
+			});
+
+			if (response.errors) {
+				throw new CustomError('graphql response contains errors', null, { response });
+			}
+		} catch (err) {
+			// handle error
```
bertyhell, line 1311: this comment doesn't add any useful information

src/collection/collection.const.ts
```diff
@@ -64,4 +64,8 @@ export const GET_MARCOM_ENTRY_TABLE_COLUMNS: () => TableColumn[] = () => [
 		label: i18n.t('collection/collection___link'),
 		id: 'external_link',
 	},
+	{
+		label: i18n.t('Acties'),
```
bertyhell, line 68: the alignment of the 'Actions' column header and the button, doesn't match (left aligned header, vs right aligned button)

In the admin panel we just use an empty string for the actions column:
```
label: '',
```

You could the same here, unless the Actions label was a requirement in the jira task?

bertyhell (approved):

PR #1214, opened 2021-06-25
fix(timebound): directly set is_public if it should be published now

bartnaessens:


src/admin/content/components/PublishContentPageModal.tsx
```diff
@@ -53,10 +53,18 @@ const PublishContentPageModal: FunctionComponent<PublishContentPageModalProps> =
...
 			const newContent: Partial<ContentPageInfo> = {
-				is_public: selectedOption === 'public',
+				is_public:
+					selectedOption === 'public' ||
+					(selectedOption === 'timebound' &&
+						((publishAt &&
+							new Date(publishAt) < now &&
+							((depublishAt && new Date(depublishAt) > now) || !depublishAt)) ||
+							(!publishAt &&
+								(!depublishAt || (depublishAt && new Date(depublishAt) > now))))),
```
bertyhell, line 65: I think this has the same effect, but is more readable. But please check my logic :)

https://lodash.com/docs/4.17.15#inRange

```
import { inRange } from 'lodash-es';

// ...

selectedOption === 'public' ||
(selectedOption === 'timebound' && 
_.inRange(
  now.getTime(), 
  publishAt ? new Date(publishAt) .getTime() : 0, 
  depublishAt ? new Date(depublishAt).getTime() : undefined
),
```

bertyhell (approved):

PR #1227, opened 2021-07-13
feat(columnfilter): add organisation column and filter

bartnaessens:
https://meemoo.atlassian.net/browse/AVO-1455

Add organisation column and filter
fixes education levels filter

Also requires: https://github.com/viaacode/avo2-proxy/pull/354

CleanShot 2021-07-13 at 11 58 39


bertyhell:
see remarks on: https://github.com/viaacode/avo2-client/pull/1228

PR #1228, opened 2021-07-13
feat(columnfilter): update organisation filter

bartnaessens:
Updated organisation filter on users page
CleanShot 2021-07-13 at 12 31 58

Based on: https://github.com/viaacode/avo2-client/pull/1227

### DIFF
--- a/src/admin/collectionsOrBundles/helpers/collection-filters.ts
+++ b/src/admin/collectionsOrBundles/helpers/collection-filters.ts
@@ -95,15 +95,27 @@ export function generateCollectionWhereObject(
 	}
 	if (filters.education_levels && filters.education_levels.length) {
 		andFilters.push({
-			_or: [
-				...getMultiOptionsFilters(
-					{
-						education_levels: without(filters.education_levels, NULL_FILTER),
-					},
-					['education_levels'],
-					['lom_context']
-				),
-			],
+			...getMultiOptionsFilters(
+				{
+					education_levels: without(filters.education_levels, NULL_FILTER),
+				},
+				['education_levels'],
+				['lom_context']
+			)[0],
+		});
+	}
+
+	if (filters.organisation && filters.organisation.length) {
+		andFilters.push({
+			...getMultiOptionsFilters(
+				{
+					organisation: filters.organisation,
+				},
+				['organisation'],
+				['owner.profile.organisation'],
+				['or_id'],
+				true
+			)[0],
 		});
 	}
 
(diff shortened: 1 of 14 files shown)


src/admin/collectionsOrBundles/helpers/collection-filters.ts
```diff
@@ -95,15 +95,27 @@ export function generateCollectionWhereObject(
...
-					['lom_context']
-				),
-			],
+			...getMultiOptionsFilters(
+				{
+					education_levels: without(filters.education_levels, NULL_FILTER),
+				},
+				['education_levels'],
+				['lom_context']
+			)[0],
```
bertyhell, line 104: there will only be one entry in the array, so you can just omit the [0]. 
since push accepts multiple entries via:
.push(...someArray);

also update the `filters.subjects` to match this notation since the subjects now still include a redundant _or statement:
```json
  {
    "_or": [
      {
        "lom_classification": {
          "_has_keys_any": [
            "aardrijkskunde",
            "autotechniek"
          ]
        }
      }
    ]
  },
```

src/admin/shared/helpers/filters.ts
```diff
@@ -128,15 +128,27 @@ export function getMultiOptionsFilters(
...
+								}
+								return {
+									[referenceTable]: { _in: value },
+								};
+							}
 							if (labelPath) {
 								return {
 									[referenceTable]: {
-										[labelPath]: { _has_key: value },
+										[labelPath]: { _has_keys_any: value },
```
bertyhell, line 146: This changes the behavior for the existing filters:

before it was an "AND" operation, now it is an "OR" operation.
For organisation this makes sense, but for subjects it might not.
You probably want to make this an option in the getMultiOptionsFilter function: operator: 'AND' | 'OR'

filter subjects on production: (AND)
image

filter subjects in this PR: (OR)
image

bertyhell (changes requested):

src/admin/collectionsOrBundles/helpers/collection-filters.ts
bartnaessens: fixed

src/admin/shared/helpers/filters.ts
bartnaessens: checked - both subjects and organizations behave as OR filters now

bartnaessens (commented):
processed feedback

bertyhell (approved):

PR #1232, opened 2021-07-19
fix(error-view): attempt to fix background-color on safari

bertyhell:
closes: https://meemoo.atlassian.net/browse/AVO-1493

requires component PR: https://github.com/viaacode/avo2-components/pull/453

Please try to access this page while logged in as a teacher/student on Safari: http://localhost:8080/beheer

You should see:
image

and not:
image

### DIFF
--- a/public/unsupported-browser.html
+++ b/public/unsupported-browser.html
@@ -16,12 +16,12 @@
 
 	body {
 		margin: 0;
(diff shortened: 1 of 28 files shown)


bartnaessens:
Not yet fixed with this PR. Upon any change the page is rendered correctly.

PR #1321, opened 2022-02-14
AVO-1727 hide entries with channel type overige

bartnaessens:
hides communication entries for collections and bundles with channel type 'overige'

### DIFF
--- a/src/admin/collectionsOrBundles/helpers/collection-filters.ts
+++ b/src/admin/collectionsOrBundles/helpers/collection-filters.ts
@@ -184,6 +184,13 @@ export function generateCollectionWhereObject(
 		},
 	});
 
+	const excludeChannelType = get(filters, 'excludeChannelType');
+	if (excludeChannelType) {
+		andFilters.push({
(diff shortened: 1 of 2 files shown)


src/admin/collectionsOrBundles/helpers/collection-filters.ts
```diff
@@ -184,6 +184,13 @@ export function generateCollectionWhereObject(
 		},
 	});
 
+	const excludeChannelType = get(filters, 'excludeChannelType');
```
bertyhell, line 187: You can use the native option for these from now on:
```
const excludeChannelType = filters?.excludeChannelType;
```

It provides better type checking.

bertyhell (approved):

src/admin/collectionsOrBundles/helpers/collection-filters.ts
bertyhell, line 187: You can use the native option for these from now on:
```suggestion
	const excludeChannelType = filters?.excludeChannelType;
```

It provides better type checking.

PR #1338, opened 2022-05-11
Bugfix/avo 1868

ian-emsens-shd:
https://meemoo.atlassian.net/browse/AVO-1868


bertyhell:
works when i run this branch on top of `master`
but when i run it on top of `epic/AVO-1740` i don't see any links in the table

bertyhell (approved):

PR #1357, opened 2022-06-17
feat(avo-1942): delete the new title input modal on duplicate action

bartnaessens:
Difficulty is re-fetch the assignments only when necessary; if you duplicate from page 2, the filters (& sort) should be reset so you see your new copy on top of the list (page 0).


src/assignment/views/AssignmentOverview.tsx
```diff
@@ -307,16 +302,21 @@ const AssignmentOverview: FunctionComponent<AssignmentOverviewProps> = ({
 					'Failed to duplicate the assignment because the marked assignment is null'
 				);
 			}
+			const newTitle = `(kopie) ${assignment.title}`;
```
ian-emsens-shd, line 305: Ik zou `(kopie)` als `i18n` string implementeren

ian-emsens-shd (approved):

PR #1379, opened 2022-07-01
feat: auto copy whitelisted query to proxy/server/scripts

bartnaessens:
set PROXY_PATH in your .env to path-to-proxy/server


ian-emsens-shd (approved):

scripts/whitelist-gql-queries.ts
```diff
@@ -3,24 +3,27 @@
...
 import * as path from 'path';
 
+import fse from 'fs-extra';
 import glob from 'glob';
 import { split } from 'lodash';
 
-const fs = require('fs-extra');
-
-require('dotenv').config({ path: path.resolve(__dirname, '../.env') });
+// require('dotenv').config({ path: path.resolve(__dirname, '../.env') });
```
bertyhell, line 10: can be deleted

bertyhell (approved):

bertyhell:
i've changed the example file to use the most likely path instead of an example path
99% of developers will checkout the proxy repo next to the client repo with names: avo2-proxy and avo2-client

I've also removed some duplicate queries

PR #1387, opened 2022-07-05
fix(avo-1910): fix render loop collection detail

bertyhell:
The render loop would occur when visiting a collection detail. eg:
http://localhost:8080/collecties/82451704-ffb0-4466-a126-4f0a1dc724de

Now it doesn't loop.

### DIFF
--- a/src/collection/views/CollectionDetail.tsx
+++ b/src/collection/views/CollectionDetail.tsx
@@ -39,7 +39,6 @@ import { PermissionName, PermissionService } from '../../authentication/helpers/
 import RegisterOrLogin from '../../authentication/views/RegisterOrLogin';
 import { APP_PATH, GENERATE_SITE_TITLE } from '../../constants';
 import { ALL_SEARCH_FILTERS, SearchFilter } from '../../search/search.const';
-import { FilterState } from '../../search/search.types';
 import {
 	IconBar,
 	InteractiveTour,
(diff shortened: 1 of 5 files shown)


src/collection/views/CollectionDetail.tsx
```diff
@@ -122,40 +121,26 @@ type CollectionDetailPermissions = Partial<{
...
-	const [collection, setCollection] = useState<Avo.Collection.Collection | null>(null);
+
+	const [collectionInfo, setCollectionInfo] = useState<{
+		collection: Avo.Collection.Collection | null;
+		permissions: CollectionDetailPermissions;
+		showLoginPopup: boolean;
+	} | null>(null);
+	const permissions = collectionInfo?.permissions;
+	const showLoginPopup = collectionInfo?.showLoginPopup;
+	const collection = collectionInfo?.collection;
```
ian-emsens-shd, line 142:
```suggestion
	const [{collection, permissions, showLoginPopup}, setCollectionInfo] = useState<{
		collection?: Avo.Collection.Collection | null;
		permissions?: CollectionDetailPermissions;
		showLoginPopup?: boolean;
	}>({});
```

ian-emsens-shd (approved):

bartnaessens (approved):

src/collection/views/CollectionDetail.tsx
bertyhell: cool

bertyhell: wait, won't that crash when the value is null, since you can't destruct null?

PR #1402, opened 2022-07-07
feat(avo-1912): make pupil collection readonly when deadline has passed

bertyhell:
https://meemoo.atlassian.net/browse/AVO-1912

image


src/assignment/views/AssignmentResponseEdit/AssignmentResponseEdit.tsx
```diff
@@ -124,8 +124,8 @@ const AssignmentResponseEdit: FunctionComponent<
 		}
 	);
 
-	const pastDeadline = useMemo(
-		() => assignment?.deadline_at && isPast(new Date(assignment.deadline_at)),
+	const pastDeadline: boolean = useMemo(
+		() => (assignment?.deadline_at && isPast(new Date(assignment.deadline_at))) || false,
 		[assignment]
 	);
```
ian-emsens-shd, line 130: Align with implementation in [`src/assignment/hooks/assignment-pupil-tabs.tsx`](https://github.com/viaacode/avo2-client/pull/1402/files#diff-a90f9148badf347235e9ccea843f8e7daec08786167408c9cac290c1787725f1) (split to hook/helper/...)

src/assignment/views/AssignmentResponseEdit/AssignmentResponseEdit.tsx
```diff
@@ -373,6 +373,7 @@ const AssignmentResponseEdit: FunctionComponent<
 				}
 				return (
 					<AssignmentResponsePupilCollectionTab
+						pastDeadline={pastDeadline}
```
ian-emsens-shd, line 376: Avoid unneeded prop after `pastDeadline` is split up.

ian-emsens-shd (changes requested):

src/assignment/views/AssignmentResponseEdit/AssignmentResponseEdit.tsx
bertyhell: AssignmentResponsePupilCollectionTab has no access to assignment, so we still need to pass the pastDeadline boolean
otherwise we need to pass the assignment object and have the component caluclate it it self with an extra hook call.

PR #1407, opened 2022-07-08
feat(avo-1912): add mobile layout for sticky save bar

bertyhell:
https://meemoo.atlassian.net/browse/AVO-1912

also noticed issue with zendesk positioning on mobile => need to investigate => separate PR

image

image


ian-emsens-shd (changes requested):
Needs work
- Avoid `isMobileWidth` to render useless elements
- Avoid using the `Flex` component entirely

The only thing that needs to happen is to set `flex-basis` to `100%` & `50%`

image

bertyhell:
> The only thing that needs to happen is to add flex-wrap and set flex-basis to 100% & 50%

Fixed

PR #1414, opened 2022-07-08
feat(add-item-coll-modal): make table scrollable

bertyhell:
so the action buttons are always visible
also make all video thumbnails the same size

before:
image
image

after:
image
image

### DIFF
--- a/src/assignment/modals/AddCollectionModal.tsx
+++ b/src/assignment/modals/AddCollectionModal.tsx
@@ -23,6 +23,7 @@ import { Avo } from '@viaa/avo2-types';
 import { noop } from 'lodash-es';
 import React, { FC, FunctionComponent, useCallback, useEffect, useMemo, useState } from 'react';
 import { useTranslation } from 'react-i18next';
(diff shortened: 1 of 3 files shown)


src/assignment/modals/AddCollectionModal.tsx
```diff
@@ -325,31 +328,34 @@ const AddCollectionModal: FunctionComponent<AddCollectionModalProps> = ({
...
+						variant="styled"
+						onColumnClick={handleColumnClick as any}
+						sortColumn={sortColumn}
+						sortOrder={sortOrder}
+						useCards={isMobileWidth()}
+						showRadioButtons
+						selectedItemIds={selectedCollectionId ? [selectedCollectionId] : []}
+						onSelectionChanged={handleSelectedCollectionChanged}
+					/>
+				</ScrollBar>
```
ian-emsens-shd, line 357: Optional; use of this component can be avoided using a max-height on the modal-body and standard scrolling

ian-emsens-shd (approved):

PR #1426, opened 2022-07-12
fix(pupil-collection): add block can only be text

bertyhell:
It seems like this one got lost during the big merging

image


bartnaessens (approved):

ian-emsens-shd (approved):

ian-emsens-shd:
Note: is `addBlockModal` still used?

PR #1428, opened 2022-07-12
fix(create-assign): fix add fragments and collections to new assignment

bartnaessens:
add callback config for modals
wrap addbookmarkfragmentmodal withUser


ian-emsens-shd (approved):
TODO; figure out how to reduce duplication between Create & Edit without locking in functionality

bertyhell:
> TODO; figure out how to reduce duplication between Create & Edit without locking in functionality

idd, mss als helper functies ofzo. Mag in aparte PR zodat we nu al kunnen deployen

PR #1448, opened 2022-07-14
fix(avo-2028): add fragment to pupil collection working draft

bertyhell:
https://meemoo.atlassian.net/browse/AVO-2028

When a pupil adds a fragment to his collection, it doesn't immediately show up under the collection tab. This is because we only added the fragment in the database and not inside the state in the client.

The state in the client can have unsaved changes, so we need to 
* add the fragment to the pupil collection in the database as the last block
* add the fragment to the assignmentResponse.pupil_collection_blocks in the client state as the last block (after the possibly unsaved blocks)

flow:
user adds a test block and types some text in the title without saving:
image

pupil goes to search tab:
image

pupil selects a fragment:
image

pupil adds fragment to collection:
image
image

pupil goes back to the collection tab:
image


src/assignment/views/AssignmentResponseEdit/AssignmentResponseEdit.tsx
```diff
@@ -207,6 +205,18 @@ const AssignmentResponseEdit: FunctionComponent<AssignmentResponseEditProps & Us
...
+	const appendBlockToPupilCollection = (newBlock: Avo.Core.BlockItemBase) => {
+		const newBlocks = setPositionToIndex([
+			...(assignmentResponse.pupil_collection_blocks || []),
+			newBlock,
+		]);
+		setAssignmentResponse({
+			...assignmentResponse,
+			pupil_collection_blocks: newBlocks,
+		});
+		setValue('pupil_collection_blocks', newBlocks as PupilCollectionFragment[]);
```
ian-emsens-shd, line 217: set dirty

src/assignment/views/AssignmentResponseEdit/AssignmentResponseEditPage.tsx
```diff
@@ -210,11 +217,22 @@ const AssignmentResponseEditPage: FunctionComponent<
 			);
 		}
 
+		if (!assignmentResponse) {
+			return (
+				<ErrorView
+					message={t('De opdracht antwoord entry kon niet worden aangemaakt')}
```
ian-emsens-shd, line 223: i18n

ian-emsens-shd (approved):

PR #1517, opened 2022-08-17
fix(avo-2121): show bookmark created date instead of item creation date

bertyhell:
https://meemoo.atlassian.net/browse/AVO-2121

before:
image

after:
image


src/assignment/modals/AddBookmarkFragmentModal.tsx
```diff
@@ -53,15 +58,15 @@ const GET_ADD_BOOKMARK_FRAGMENT_COLUMNS = (): TableColumn[] => [
 		dataType: 'string',
 	},
 	{
-		id: 'contentCreatedAt',
+		id: 'createdAt',
```
ian-emsens-shd, line 61:
```suggestion
		id: AddBookmarkFragmentColumn.createdAt,
```

Also applies to other `id`'s

ian-emsens-shd (approved):

PR #1519, opened 2022-08-17
fix(avo-2116): avoid showing 2 interactive tour buttons

bertyhell:
https://meemoo.atlassian.net/browse/AVO-2116

before:
image

after:
image

the tour button on the item detail page still works as well:
image


ian-emsens-shd (changes requested):
Keep it consistent with `renderBookmarkCount={() => null}`

ian-emsens-shd (approved):

PR #1525, opened 2022-08-19
fix(avo-1841): delete cover image during content page duplication

bertyhell:
https://meemoo.atlassian.net/browse/AVO-1841

also made sure the delete confirm modal is closed when confirming.
This behavior was changed some time ago,
and I noticed some modals do not close correctly after a successful delete

duplicate a page with a cover image:
image

the duplicated page should now no longer have an image:
image

delete modal now correctly close after confirming:
image


src/admin/content/content.service.ts
```diff
@@ -525,43 +524,44 @@ export class ContentService {
 		profileId: string
 	): Promise<Partial<ContentPageInfo> | null> {
 		try {
-			const contentToInsert = { ...contentPageInfo };
+			const duplicatedContentPage = { ...contentPageInfo };
```
ian-emsens-shd, line 527: Aangezien de functie (`duplicateContentPage`) reeds de context van de entiteit (`ContentPage`) mee geeft ben ik eerder voorstander om dit soort variabelen kort te houden op bv. [the] `const duplicate = ...`.

ian-emsens-shd (approved):

PR #1542, opened 2022-09-05
Bugfix/avo 2085

ian-emsens-shd:
https://meemoo.atlassian.net/browse/AVO-2085

Requires https://github.com/viaacode/avo2-proxy/pull/492

Note: this PR also fixes inconsistent metadata padding


bertyhell:
i created a pupil collection with 3 items:
image

I depublished the first without reason,
i depublished the second with a reason,
i depublished the third with a replacement:

This has the desired result:
image

But when i republish all 3 items i see this:
image


Shouldn't the last item show the original again? or should it keep showing the replaced item?

src/admin/items/items.service.ts
```diff
@@ -258,63 +259,80 @@ export class ItemsService {
 		);
 	}
 
-	public static async fetchItemByExternalId(
-		externalId: string
-	): Promise<(Avo.Item.Item & { replacement_for?: string }) | null> {
+	public static async fetchItemsByExternalId(
+		externalIds: string[]
+	): Promise<Array<(Avo.Item.Item & { replacement_for?: string }) | null>> {
```
bertyhell, line 264: null should not be a possible response type here, since you always return at least an empty array.
If you do sometimes return null, you need to handle it correctly in the code that uses this function.

src/admin/items/items.service.ts
```diff
@@ -258,63 +259,80 @@ export class ItemsService {
...
-				externalId,
-				query: 'GET_ITEM_BY_EXTERNAL_ID',
+			throw new CustomError('Failed to get items by external id', err, {
+				externalIds,
+				query: 'GET_ITEMS_BY_EXTERNAL_ID',
 			});
 		}
 	}
 
+	public static async fetchItemReplacementByExternalId(externalId: string) {
```
bertyhell, line 303:
```suggestion
	public static async fetchItemReplacementByExternalId(externalId: string): Promise<Avo.Item.Item | null> {
```

src/assignment/views/AssignmentResponses.tsx
```diff
@@ -243,8 +258,18 @@ const AssignmentResponses: FunctionComponent<AssignmentResponsesProps> = ({
...
 			);
 
+			// Determine each response's fragments to evaluate their published-status
+			const fragments = response.assignmentResponses
+				.flatMap((response) =>
+					((response.pupil_collection_blocks as PupilCollectionFragment[]) || []).flatMap(
+						(block) => block.fragment_id
+					)
+				)
+				.filter((value, index, self) => self.indexOf(value) === index);
```
bertyhell, line 268: what does this do?

src/admin/items/items.service.ts
```diff
@@ -258,63 +259,80 @@ export class ItemsService {
...
 		} catch (err) {
-			throw new CustomError('Failed to get item by external id', err, {
-				externalId,
-				query: 'GET_ITEM_BY_EXTERNAL_ID',
+			throw new CustomError('Failed to get items by external id', err, {
+				externalIds,
+				query: 'GET_ITEMS_BY_EXTERNAL_ID',
 			});
 		}
 	}
```
bertyhell, line 301: optional:
```suggestion
	}
	
	public static async fetchItemByExternalId(
		externalId: string
	): Promise<(Avo.Item.Item & { replacement_for?: string }) | null> {
		return (await ItemsService.fetchItemByExternalId([externalId]))[0] || null;
	}
```

bertyhell (approved):

src/admin/items/items.service.ts
ian-emsens-shd: Items in the array can be `null`, not the entire response.
`Promise<Array<... | null>>`

src/assignment/views/AssignmentResponses.tsx
ian-emsens-shd: Prevents duplicate string values in the payload sent when querying for items.

ian-emsens-shd:
> Shouldn't the last item show the original again? or should it keep showing the replaced item?

Items are fetched based on the existing logic in the item service; what happens to the originally replaced item? When I add QAS's `5d8nc92783` to a pupil collection the original item shows up correctly.

src/assignment/views/AssignmentResponses.tsx
bertyhell: mss is lodash uniq of uniqBy dan een duidelijkere oplossing:
https://lodash.com/docs/4.17.15#uniqBy

PR #1543, opened 2022-09-05
fix(avo-2140): show label column in assign. overview for pupil & teacher

bertyhell:
https://meemoo.atlassian.net/browse/AVO-2140

after:

### DIFF
--- a/src/assignment/views/AssignmentOverview.tsx
+++ b/src/assignment/views/AssignmentOverview.tsx
@@ -500,16 +500,7 @@ const AssignmentOverview: FunctionComponent<AssignmentOverviewProps> = ({
 			}
 
 			case 'labels':
-				return labels.length > 0 ? (
-					<>
-						{renderLabels(labels)}
-						{renderDataCell(
-							labels.map((label) => label.assignment_label.label).join(', '),
-							`${t('assignment/views/assignment-overview___labels')}`,
-							'm-assignment-overview__table__data-cell--labels'
-						)}
-					</>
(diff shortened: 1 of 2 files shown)


src/assignment/views/AssignmentOverview.tsx
```diff
@@ -500,16 +500,7 @@ const AssignmentOverview: FunctionComponent<AssignmentOverviewProps> = ({
...
 
 			case 'labels':
-				return labels.length > 0 ? (
-					<>
-						{renderLabels(labels)}
-						{renderDataCell(
-							labels.map((label) => label.assignment_label.label).join(', '),
-							`${t('assignment/views/assignment-overview___labels')}`,
-							'm-assignment-overview__table__data-cell--labels'
-						)}
```
ian-emsens-shd, line 510: This is used for mobile layout, should remain intact.
see `src/assignment/views/AssignmentOverview.scss`

ian-emsens-shd (changes requested):

src/assignment/views/AssignmentOverview.tsx
bertyhell: fixed: 
image

ian-emsens-shd (changes requested):
Maintaining two instances of DOM is a lot more performant than using `isMobileWidth` to only output one. Restore the media query and avoid using `isMobileWidth` in the future.

ian-emsens-shd (approved):

PR #1544, opened 2022-09-05
Feature/avo 1833

ian-emsens-shd:
https://meemoo.atlassian.net/browse/AVO-1833


bertyhell (approved):

src/settings/components/LinkedAccounts.tsx
```diff
@@ -205,10 +212,8 @@ const LinkedAccounts: FunctionComponent<AccountProps> = ({ location, user }) =>
 										'settings/components/account___koppel-je-account-met-andere-platformen'
 									)}
 								>
-									{/* Disabled for pupils due to https://meemoo.atlassian.net/browse/AVO-2062 */}
-									<div>
-										{!isPupil && renderIdpLinkControls('VLAAMSEOVERHEID')}
-									</div>
+									{/* Previously disabled for pupils due to https://meemoo.atlassian.net/browse/AVO-2062 */}
```
bertyhell, line 215: do we still need this?

src/settings/components/LinkedAccounts.tsx
```diff
@@ -205,10 +212,8 @@ const LinkedAccounts: FunctionComponent<AccountProps> = ({ location, user }) =>
...
+									<div>{renderIdpLinkControls('VLAAMSEOVERHEID')}</div>
```
bertyhell, line 216: button is not the same size on mobile:
image

bertyhell (changes requested):

src/settings/components/LinkedAccounts.tsx
ian-emsens-shd: I'd prefer to keep the record here

bertyhell:
The link screen is quite confusing now.
image


Would we change it to this design instead?
image

PR #1548, opened 2022-09-06
Bugfix/avo 2143

ian-emsens-shd:
https://meemoo.atlassian.net/browse/AVO-2143

Tested locally with a collection of 30 `ITEM`-blocks, significant improvement but there's still noticeable lag `onBlur`.


bertyhell (approved):

bertyhell:
i don't see much difference, but for me it didn't take 4 seconds before this change. So we'll need to see how it does on a slower machine.

PR #1580, opened 2022-09-29
feat(avo-1853): process account link query param and show message

ian-emsens-shd:
https://meemoo.atlassian.net/browse/AVO-1853

Requires https://github.com/viaacode/avo2-proxy/pull/506


src/App.tsx
```diff
@@ -127,6 +127,21 @@ const Root: FunctionComponent = () => {
...
+		((key: Avo.Auth.IdpLinkedSuccessQueryParam = 'linked') => {
+			const hasLinked = url.searchParams.get(key) !== null;
+
+			if (hasLinked) {
+				ToastService.success(t('app___je-account-is-gekoppeld'));
+				url.searchParams.delete(key);
+				history.replace(url.toString().replace(url.origin, ''));
+			}
+		})();
+	}, []);
```
bertyhell, line 143: This is a strange way to enforce a type
Why not just define it as a typed variable, then the typescript compiler will complain if it isn't the correct type:
image

```suggestion
	useEffect(() => {
		const url = new URL(window.location.href);
		const linked: Avo.Auth.IdpLinkedSuccessQueryParam = 'linked';
		const hasLinked = url.searchParams.get(linked) !== null;
		if (hasLinked) {
			ToastService.success(t('app___je-account-is-gekoppeld'));
			url.searchParams.delete(linked);
			history.replace(url.toString().replace(url.origin, ''));
		}
	}, []);
```

bertyhell (changes requested):

PR #1582, opened 2022-09-30
fix(avo-2136): rework nudging modal show/hide

bertyhell:
Currently the nudging modal is visible after logging in with itsme.
That is probably due to:
```
			const hasVlaamseOverheidLinked =
				!!user &&
				hasIdpLinked(user, 'VLAAMSEOVERHEID__SUB_ID') &&
				hasIdpLinked(user, 'VLAAMSEOVERHEID__ACCOUNT_ID');
```
which should be using an OR operator:
```			
			const hasVlaamseOverheidLinked =
				!!user &&
				(hasIdpLinked(user, 'VLAAMSEOVERHEID__SUB_ID') ||
					hasIdpLinked(user, 'VLAAMSEOVERHEID__ACCOUNT_ID'));
```

But also this doesn't seem to react to changes in routes, since we only update the show/hide modal if certain conditions are true. But we should also hide it, in the other case when the conditions are false.
```
		if (user && !isOnAccountLinkingPage && (!isPupil || (isPupil && !isOnAssignmentPage))) {
			fetchProfilePreference();
		}
```

### DIFF
--- a/src/shared/components/ACMIDMNudgeModal/ACMIDMNudgeModal.tsx
+++ b/src/shared/components/ACMIDMNudgeModal/ACMIDMNudgeModal.tsx
@@ -18,7 +18,7 @@ import { NOT_NOW_LOCAL_STORAGE_KEY, NOT_NOW_VAL, ROUTE_PARTS } from '../../const
 import { CustomError } from '../../helpers';
 import withUser, { UserProps } from '../../hocs/withUser';
 import {
-	ProfilePreference,
+	ProfilePreferenceKey,
 	ProfilePreferencesService,
 } from '../../services/profile-preferences.service';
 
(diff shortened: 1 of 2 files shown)


src/shared/components/ACMIDMNudgeModal/ACMIDMNudgeModal.tsx
```diff
@@ -94,9 +68,29 @@ const ACMIDMNudgeModal: FC<UserProps & UiStateProps & RouteComponentProps> = ({
...
+			);
+		} else {
+			setShowNudgingModal(false);
 		}
-	}, [fetchProfilePreference, user, isPupil, location, setShowNudgingModal]);
+	}, [user, isPupil, location, setShowNudgingModal]);
+
+	useEffect(() => {
+		updateShowNudgingModal();
+	}, [updateShowNudgingModal]);
```
ian-emsens-shd, line 93: This is functionally identical to using one effect because the dependency array`[updateShowNudgingModal]` equals `[user, isPupil, location, setShowNudgingModal]`.

ian-emsens-shd (approved):

src/shared/components/ACMIDMNudgeModal/ACMIDMNudgeModal.tsx
bertyhell: yes, but then you can't use await inside the useEffect, because it doesn't accept a handler that returns a promise

PR #1586, opened 2022-10-03
feat(avo-1924): add blue save bar to collection edit

ian-emsens-shd:
https://meemoo.atlassian.net/browse/AVO-1924

### DIFF
--- a/src/collection/components/CollectionOrBundleEdit.tsx
+++ b/src/collection/components/CollectionOrBundleEdit.tsx
@@ -10,9 +10,6 @@ import {
 	Spacer,
 	TabProps,
 	Tabs,
-	Toolbar,
-	ToolbarItem,
-	ToolbarLeft,
 } from '@viaa/avo2-components';
 import { Avo } from '@viaa/avo2-types';
 import { cloneDeep, get, isEmpty, set } from 'lodash-es';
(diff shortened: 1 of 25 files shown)


src/collection/components/CollectionOrBundleEdit.tsx
```diff
@@ -1193,25 +1208,27 @@ const CollectionOrBundleEdit: FunctionComponent<
...
 						<Tabs tabs={tabs} onClick={selectTab} />
 					</Container>
 				</Navbar>
-				{renderTab()}
-				<Container background="alt" mode="vertical">
-					<Container mode="horizontal">
-						<Toolbar autoHeight>
-							<ToolbarLeft>
-								<ToolbarItem>
-									<ButtonToolbar>{renderSaveButton()}</ButtonToolbar>
```
bertyhell, line 1209: shouldn't the other save button also be removed?
image

bertyhell (changes requested):

src/collection/components/CollectionOrBundleEdit.tsx
ian-emsens-shd: Not in my opinion, ticket only mentions adding the blue bar and doesn't even mention removing the save button at the bottom. It's also useful for existing users, especially here, to keep the button at the top as they've been using it for years.

ian-emsens-shd: https://studiohyperdrive.slack.com/archives/CJ55B3Q2W/p1664991960911719?thread_ts=1664978539.792149&cid=CJ55B3Q2W

PR #1591, opened 2022-10-05
fix(avo-1813): adjust source for temp_access

ian-emsens-shd:
https://meemoo.atlassian.net/browse/AVO-1813

proxy changes: https://github.com/viaacode/avo2-proxy/pull/510


src/settings/components/Profile.tsx
```diff
@@ -660,9 +660,9 @@ const Profile: FunctionComponent<
 			}
 
 			case 'temp_access': {
-				const tempAccess = get(user, 'temp_access.current.status');
+				const tempAccess = get(profile, 'user.temp_access');
 
-				return tempAccess === 1
```
bertyhell, line 665: this shows the correct value in the table, but didn't you also say something about, deleting the temp access entry in the database when someone gets permenant access?

Where is the code that does that deleting?

bertyhell (commented):
see code remark

bertyhell (approved):

PR #1602, opened 2022-10-07
INFRA-623 add limits

violetina:
test build location on specific role of node and reduce max memory for buld , add pod limits


Dockerfile
```diff
@@ -11,32 +10,29 @@ ENV CI $CI
...
 RUN mkdir ./build/ &&chown -R node:node /app && chmod -R  g+s /app && chmod -R  g+w /app
-COPY  . .
+COPY  --chown=node:node . .
 RUN chown -R node:node /app && chmod -R  g+sw /app
 RUN apk update
 RUN apk add --no-cache --virtual python2 make g++ tzdata && ln -snf /usr/share/zoneinfo/$TZ /etc/localtime && echo $TZ > /etc/timezone
-#USER node
-RUN npm ci --production=false --force --legacy-peer-deps
+USER node
+RUN npm ci --include=dev --legacy-peer-deps
```
bertyhell, line 18: to be able to run the build, we need to dev dependencies
so the --production=false option will be needed

PR #1610, opened 2022-10-13
Feature/avo 1885

ian-emsens-shd:
https://meemoo.atlassian.net/browse/AVO-1885


src/workspace/views/QuickLaneOverview.tsx
```diff
@@ -215,38 +247,134 @@ const QuickLaneOverview: FunctionComponent<QuickLaneOverviewProps> = ({ user })
...
+								if (selected === undefined) {
+									return;
+								}
+
+								switch (action.toString() as actions) {
+									case 'edit':
+										setSelected(
+											await QuickLaneService.fetchQuickLaneById(selected.id)
+										);
+										setIsQuickLaneModalOpen(true);
```
bertyhell, line 297: if the collection to which the quicklane points does not exist, then the fetch will fail, but the user does not see any error toast. it just fails silently:
image
image

going to the quicklane link shows this screen:
image
I would expect to see: this collection no longer exists, or something similar.

bertyhell (changes requested):
see code remark

bertyhell (approved):

PR #1613, opened 2022-10-13
Feature/avo 2164

ian-emsens-shd:
https://meemoo.atlassian.net/browse/AVO-2164


bertyhell:
i also like opening a pr directly on the client to update the typings or the avo2-components
but a link to the avo2-components commit where you made the changes, helps speed up the review process

bertyhell (approved):

PR #1618, opened 2022-11-02
feat(avo-2181): remove excess bookmarks

ian-emsens-shd:
https://meemoo.atlassian.net/browse/AVO-2181?focusedCommentId=35540

Requires https://github.com/viaacode/avo2-proxy/pull/530

### DIFF
--- a/src/admin/items/items.gql.ts
+++ b/src/admin/items/items.gql.ts
@@ -360,6 +360,11 @@ export const REPLACE_ITEM_IN_COLLECTIONS_BOOKMARKS_AND_ASSIGNMENTS = gql`
 		) {
 			affected_rows
 		}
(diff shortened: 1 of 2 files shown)


bertyhell:
image


looks like the new version of hasura is causing an issue with the id argument
the id should be uuid

image


but it isn't used in the query, which also isn't allowed anymore in the latest version of hasura
I suspect the id was passed to enable permission checking to make sure those fragments 

We can probably change the check to check the collection uuid in the fragments itself.
We might also want to add a check if the array of fragments is empty, to not do the network call

I'll check with Christophe if we can fix issues like this on support.

bertyhell (changes requested):
see global remark

bertyhell (approved):

PR #1620, opened 2022-11-03
fix(avo-1730): bump components

ian-emsens-shd:
https://meemoo.atlassian.net/browse/AVO-1730

https://github.com/viaacode/avo2-components/compare/3e2c78aaa1c23c816bb433cc60ebcd5f233a7d74...d64d1d24070acc41958f67f37b47173216f14564

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,7 +14,7 @@
         "@popperjs/core": "^2.5.4",
         "@react-hook/resize-observer": "^1.2.5",
         "@storybook/addon-actions": "^6.3.12",
-        "@viaa/avo2-components": "^1.97.17",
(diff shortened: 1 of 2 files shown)


bertyhell:
why do we only hide the "video", "collection", "audio" labels and not remove the html altogether?
is that for accessibility?
```
			p {
				display: none;
			}
```

ian-emsens-shd:
> why do we only hide the "video", "collection", "audio" labels and not remove the html altogether? is that for accessibility?
> 
> ```
> 			p {
> 				display: none;
> 			}
> ```

Labels need to remain on other pages, they're just hidden in a certain block. No use polluting the component with render functions or boolean props to avoid rendering non-interactable DOM.

PR #1621, opened 2022-11-03
fix(avo-2219): adjust flex layout in modal

ian-emsens-shd:
https://meemoo.atlassian.net/browse/AVO-2219?focusedCommentId=35412


bertyhell:
the mobile view of the modal is still fucked up:
image

bertyhell (changes requested):
see global remark

PR #1625, opened 2022-11-03
feat(avo-2211): add permissions overview and admin config for admin-core

bertyhell:
This PR fixes these issues:
* https://meemoo.atlassian.net/browse/AVO-2197
* https://meemoo.atlassian.net/browse/AVO-2198

And partly resolves:
* https://meemoo.atlassian.net/browse/AVO-2211

Requires proxy PR: https://github.com/viaacode/avo2-proxy/pull/528

Still todo: styling of the permissions screen

image


src/App.tsx
```diff
@@ -140,6 +140,7 @@ const Root: FunctionComponent = () => {
 		}
 	}, []);
 
+	console.log('setting query client provider, ', queryClient);
```
ian-emsens-shd, line 143: Wrap in an environment check or remove

src/admin/admin.const.ts
```diff
@@ -75,22 +73,13 @@ function getUserNavItems(userPermissions: string[]): NavigationItemInfo[] {
 			},
 			{
 				navItem: {
-					label: i18n.t('admin/admin___gebruikersgroepen'),
+					label: i18n.t('Groepen en permissies'),
```
ian-emsens-shd, line 76: `$ npm run extract-translations`

src/admin/shared/hoc/with-admin-core-config.tsx
```diff
@@ -0,0 +1,228 @@
...
+import { UserGroupsService } from '../services/user-groups';
+
+const InternalLink = (linkInfo: LinkInfo) => {
+	return <Link {...linkInfo} />;
+};
+
+export const withAdminCoreConfig = (WrappedComponent: ComponentType): ComponentType => {
+	const Component = (props: Record<string, unknown>) => {
+		const [adminCoreConfig, setAdminCoreConfig] = useState<AdminConfig | null>(null);
+		const user = props.user as Avo.User.User | undefined;
```
ian-emsens-shd, line 38: I'd prefer to avoid this cast and make props more strongly typed
```
props: {
  user?: Avo.User.User,
  [key: string]: unknown
}
```
Ideally we can infer the rest of the props from `ComponentType` but that might require a lot of dynamic typing.

src/admin/shared/hoc/with-admin-core-config.tsx
```diff
@@ -0,0 +1,228 @@
...
+			};
+
+			const config: AdminConfig = {
+				navigation: {
+					service: navigationService,
+					views: {
+						overview: {
+							labels: { tableHeads: {} },
+						},
+					},
```
ian-emsens-shd, line 68: Does this empty config need to be here? Doesn't the admin core provide fallback / handle nulls?

src/admin/shared/hoc/with-admin-core-config.tsx
```diff
@@ -0,0 +1,228 @@
...
+						ContentBlockType.PageOverview,
+						ContentBlockType.UspGrid,
+						ContentBlockType.Quote,
+					],
+					defaultPageWidth: 'LARGE',
+					onSaveContentPage: noop,
+				},
+				navigationBars: { enableIcons: false },
+				icon: {
+					component: ({ name }: { name: IconName }) => <Icon name={name} />,
```
ian-emsens-shd, line 97: Can't we just use `component: Icon` here? This looks a lot like a minimal FC.

src/admin/shared/hoc/with-admin-core-config.tsx
```diff
@@ -0,0 +1,228 @@
...
+				},
+				firstName: user?.first_name || undefined,
+				lastName: user?.last_name || undefined,
+				fullName: user?.full_name || undefined,
+				last_access_at: user?.last_access_at || undefined, // TODO enable once last_access_at field is added to the database
+				// eslint-disable-next-line @typescript-eslint/no-explicit-any
+				permissions: user?.profile?.permissions as any[],
+			};
+
+			const config: AdminConfig = {
```
ian-emsens-shd, line 61: In general I'd move the static parts of this config into a dedicated ts file to improve formatting and legibility.

src/admin/shared/services/api-service/api.service.ts
```diff
@@ -0,0 +1,32 @@
...
+				hooks: {
+					afterResponse: [
+						(_request, _options, response) => {
+							if (response.status === 401 && !ignoreAuthError) {
+								// User is no longer logged in => force them to login again
+								goToLoginBecauseOfUnauthorizedError();
+							}
+						},
+					],
+				},
```
ian-emsens-shd, line 27: Does this conflict with any existing mechanisms (i.e. `check-login`) or make them obsolete?

src/admin/user-groups/views/UserGroupOverviewPage.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { GENERATE_SITE_TITLE } from '../../../constants';
+import { withAdminCoreConfig } from '../../shared/hoc/with-admin-core-config';
+import { AdminLayout, AdminLayoutBody } from '../../shared/layouts';
+
+const UserGroupGroupOverviewPage: FunctionComponent = () => {
+	const [t] = useTranslation();
+
+	const renderPageContent = () => {
+		return <UserGroupOverview />;
+	};
```
ian-emsens-shd, line 15: What's the benefit of this function?

ian-emsens-shd (approved):

src/admin/shared/hoc/with-admin-core-config.tsx
bertyhell: i don't think these are used even. Probably legacy from when Andry was going to add all the navigation sidebars in the admin-core as well.

src/admin/shared/hoc/with-admin-core-config.tsx
bertyhell: No these are different icon components:
* react-components
* avo2-components

The client decides which library to use, so the icons match with the theme of the client.

src/admin/shared/services/api-service/api.service.ts
bertyhell: in principle, we'd like to use this service for all network calls.
Currently the dataService and fetchWithLogout function do something similar.

src/admin/user-groups/views/UserGroupOverviewPage.tsx
bertyhell: This was just to get all components uniform. They all have a layout with <MetaTags> and a {renderPageContent()}
But in this case, it doesn't really add much.

Do you prefer uniformity or simplicity? :p

PR #1633, opened 2022-11-09
Feat AVO-2211 Add styling and fix save for user group permissions overview

bertyhell:
https://meemoo.atlassian.net/browse/AVO-2211

Requires Admin Core PR: https://github.com/viaacode/react-admin-core-module/pull/57

save buttons:
image

success toast:
image

search:
image

scrolling table headers:
image


src/admin/user-groups/views/UserGroupOverviewPage.scss
```diff
@@ -0,0 +1,108 @@
...
+		}
+
+		th:first-child,
+		td:first-child {
+			width: 25%;
+		}
+
+		th:not(:first-child),
+		td:not(:first-child) {
+			width: calc(75% / 13);
```
ian-emsens-shd, line 93: Not very scalable for future groups, do we want to leave this static?

ian-emsens-shd (approved):

src/admin/user-groups/views/UserGroupOverviewPage.scss
ian-emsens-shd: Actually unneeded, I'll remove it

PR #1635, opened 2022-11-09
feat(avo-2199): Add content-page module and overview from core

ian-emsens-shd:
https://meemoo.atlassian.net/browse/AVO-2199

http://localhost:8080/beheer/core/content created alongside http://localhost:8080/beheer/content as a separate module to allow for side-by-side comparison. I'd opt to disable the old routes and wait to remove the dead code in bulk before release.

Pairs with https://github.com/viaacode/react-admin-core-module/pull/58/files
⚠️ Requires the above PR to be merged, released and its new version added to package.json


bertyhell:
> http://localhost:8080/beheer/core/content created alongside http://localhost:8080/beheer/content as a separate module to allow for side-by-side comparison. I'd opt to disable the old routes and wait to remove the dead code in bulk before release.

We can always compare to QAS or the master branch
changing routes just before the release is a bit risky. I'd prefer that we already test the final configuration while were developing all the features.

WardVerc1 (approved):

bertyhell (approved):

PR #1653, opened 2022-11-24
fix(AVO-2239): replaced flowplayer with same player as in AddToCollec…

WardVerc1:
…tion

### DIFF
--- a/src/collection/components/modals/CutFragmentModal.tsx
+++ b/src/collection/components/modals/CutFragmentModal.tsx
@@ -161,7 +161,10 @@ const CutFragmentModal: FunctionComponent<CutFragmentModalProps> = ({
 					showTitle
 					showDescription={false}
 					canPlay={isOpen}
-					cuePoints={{ start, end }}
+					cuePointsLabel={{
+						start: fragmentStartTime,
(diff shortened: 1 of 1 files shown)


src/collection/components/modals/CutFragmentModal.tsx
```diff
@@ -150,22 +150,19 @@ const CutFragmentModal: FunctionComponent<CutFragmentModalProps> = ({
...
-					cuePointsLabel={{
-						start: fragmentStartTime,
-						end: fragmentEndTime,
-					}}
+				<ItemVideoDescription
+					itemMetaData={itemMetaData}
+					showTitle
+					showDescription={false}
 					canPlay={isOpen}
+					cuePoints={{ start, end }}
```
bertyhell, line 164: Since the old code used these values, i'm assuming the changed code should use the same values.
```suggestion
					cuePoints={{ 
						start: fragmentStartTime,
						end: fragmentEndTime,
					}}
```

bertyhell (changes requested):
see code remark

ian-emsens-shd (approved):

PR #1657, opened 2022-12-01
Feature/avo 2205

WardVerc1:
https://meemoo.atlassian.net/browse/AVO-2205


bertyhell:
I made a few tweaks to the styling to better match the other tables and modals in avo:

before:
image
image

after:
image
image

bertyhell (approved):

package.json
```diff
@@ -30,12 +30,13 @@
   "dependencies": {
     "@flowplayer/player": "3.2.8",
     "@hookform/resolvers": "^2.9.9",
+    "@meemoo/admin-core-ui": "v2.0.8",
```
ian-emsens-shd, line 33: Why downgrade this dependency?

src/admin/translations/views/TranslationsOverviewV2.module.scss
```diff
@@ -0,0 +1,154 @@
...
+			width: 100%;
+			appearance: none;
+			transition: all 0.15s ease-in-out;
+
+			&:focus {
+				outline: 0;
+				border-color: #25a4cf;
+				box-shadow: inset 0 0 0 0.1rem #25a4cf;
+			}
+		}
```
ian-emsens-shd, line 40: Most of these colors are probably available in the variables.

edit: check other values for occurence in variables as well

ian-emsens-shd (changes requested):

src/admin/translations/views/TranslationsOverviewV2.module.scss
bertyhell: fixed

PR #1658, opened 2022-12-01
feat(AVO-2202): use NavigationOverview from admin-core

WardVerc1:


src/admin/menu/views/MenuOverview.tsx
```diff
@@ -1,115 +1,21 @@
...
-				data={data}
-				renderCell={(rowData: Partial<Avo.Menu.Menu>, columnId: string) =>
-					renderTableCell(rowData, columnId as MenuOverviewTableCols)
-				}
-				rowKey="id"
-				variant="bordered"
-			/>
-		);
-	};
+	const [menus] = useState<any>([]);
```
ian-emsens-shd, line 18: Is this state still used? If it is, improve the typing.

ian-emsens-shd (changes requested):

src/admin/menu/views/MenuOverview.tsx
bertyhell: I think the top bar will need to be moved to the NavigationOverview component in admin-core-ui. So that the add button can be shown depending on the navigationBars that were fetched inside the NavigationOverview component
```
			{!!menus.length && (
				<AdminLayoutTopBarRight>
					<Button
						label={tText('admin/menu/views/menu-overview___navigatie-toevoegen')}
						onClick={() => history.push(MENU_PATH.MENU_CREATE)}
					/>
				</AdminLayoutTopBarRight>
			)}
```

bertyhell (changes requested):
see code remarks

PR #1660, opened 2022-12-06
Feature/avo 2203: Added NavigationDetail from admin-core

WardVerc1:


ian-emsens-shd (changes requested):
Improve scss by using more established variables

src/admin/menu/views/MenuDetail.scss
```diff
@@ -29,3 +35,26 @@
 .c-menu-detail__nav-label {
 	margin: 0 auto 0 ($g-spacer-unit * 3);
 }
+
+.c-admin__header {
+	border-bottom: 0.1rem solid #d6dee3;
+	background-color: #edeff2;
```
bertyhell, line 41: Search for the hard coded colors in the colors.scss file and replace them with variable names. Example:
```
	@import '../../../../styles/settings/colors';

	...

	background-color: $color-gray-50;
```

src/admin/menu/views/MenuDetail.scss
```diff
@@ -29,3 +35,26 @@
...
+	display: flex;
+	padding: 20px 40px;
+
+	.c-admin__page-title {
+		font-size: 2.4rem;
+		font-family: 'Avenir LT Std', 'Avenir Next', Avenir, Montserrat, sans-serif;
```
bertyhell, line 47: you can use a variable for this:
```
@import '../../../styles/settings/variables';

...

font-family: $g-headings-font-family;
```

src/admin/menu/views/MenuDetail.scss
```diff
@@ -29,3 +35,26 @@
...
+		font-weight: 500;
+	}
+}
+
+.c-admin__content {
+	margin: 40px;
```
bertyhell, line 53: always use rem units. They usually are px value divided by 10. Rem units offer better support for people that have poor vision since they can set their browser to make rem units bigger and easier to read.
```suggestion
	margin: 4rem;
```

bertyhell:
You can make reviewing these PR's that build on top of the last one easier by setting the PR to merge into the previous branch instead of release/v1.20

You can modify that at the top of the PR:
image

This makes it so that the code that needs to be reviewed in this PR will only contain the changes for MenuDetail and not both the changes for MenuOverview and MenuDetail.

When we merge the MenuOverview PR, github will automagically change the target for this PR to release/v1.20 :D

PR #1673, opened 2023-01-05
fix(avo-2239): set player start point when initializing + select visible

bertyhell:
Sometimes the cut video modal can get into a state where the range control does not update the video.
https://user-images.githubusercontent.com/1710840/210817760-917aad1a-355a-43a1-ab27-7c5648839e25.mp4

This is because there can be 2 videos in different modals, and we're updating the wrong video in a hidden modal:
image
image

Also when you move the range control, the video does not jump to that location after the video initializes.
So i added an onPlay once handler to set the video start point to the start cuepoint when the video is initialized.

### DIFF
--- a/src/collection/components/modals/CutFragmentModal.tsx
+++ b/src/collection/components/modals/CutFragmentModal.tsx
@@ -8,7 +8,7 @@ import {
 	ToolbarRight,
 } from '@viaa/avo2-components';
 import { Avo } from '@viaa/avo2-types';
-import { noop } from 'lodash-es';
+import { noop, once } from 'lodash-es';
 import React, { FunctionComponent, useState } from 'react';
 import { useTranslation } from 'react-i18next';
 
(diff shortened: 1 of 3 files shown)


src/collection/components/modals/CutFragmentModal.tsx
```diff
@@ -145,6 +145,10 @@ const CutFragmentModal: FunctionComponent<CutFragmentModalProps> = ({
 		onClose();
 	};
 
+	const startStartTimeOnce = once(() => {
+		setModalVideoSeekTime(fragmentStartTime);
+	});
```
WardVerc1, line 150: Maybe extract and reuse this function by both components?

bertyhell: yes, i could extract this:
```
export const setModalSeekTimeOnce = () => once(() => {
		setModalVideoSeekTime(fragmentStartTime);
});
```

But then i still need to add the const in the component and import the function:
```
import { setModalSeekTimeOnce } from '...helpers';

const startStartTimeOnce = setModalSeekTimeOnce();
```

So this solution would be 7 lines of code
While the current solution is 6
So it isn't really worth it. but if the function was a little bigger, i would do it. 👍

PR #1678, opened 2023-01-12
feat(AVO-2344): remove alias, make profile fields readonly

WardVerc1:
image


src/settings/components/Profile.tsx
```diff
@@ -100,9 +101,9 @@ const Profile: FunctionComponent<
 	const [selectedOrganisations, setSelectedOrganisations] = useState<
 		ClientEducationOrganization[]
 	>(get(user, 'profile.organizations', []));
-	const [firstName, setFirstName] = useState<string>(get(user, 'first_name') || '');
-	const [lastName, setLastName] = useState<string>(get(user, 'last_name') || '');
-	const [alias, setAlias] = useState<string>(user ? getProfileAlias(user) : '');
+	const [firstName] = useState<string>(get(user, 'first_name') || '');
+	const [lastName] = useState<string>(get(user, 'last_name') || '');
+	const [email] = useState<string>(get(user, 'mail') || '');
```
silkederudder-shd, line 106: I think we can remove the `useState` and simply store the values in a const. The empty fallback string kan also be set in the lodash get function.

```suggestion
	const firstName = get(user, 'first_name', '');
	const lastName = get(user, 'last_name, '');
	const email = get(user, 'mail, '');
```

src/settings/components/Profile.tsx
```diff
@@ -681,43 +684,34 @@ const Profile: FunctionComponent<
...
 										label={t('settings/components/account___achternaam')}
 										labelFor="last_name"
-										error={get(profileErrors, 'last_name')}
 									>
-										<TextInput
-											id="last_name"
-											value={lastName || ''}
-											onChange={setLastName}
-										/>
+										{lastName || ''}
```
silkederudder-shd, line 706: Same here ;)

src/settings/components/Profile.tsx
```diff
@@ -681,43 +684,34 @@ const Profile: FunctionComponent<
...
 											>
-												<TextInput
-													id="alias"
-													placeholder={t(
-														'settings/components/profile___een-unieke-gebruikersnaam'
-													)}
-													value={alias || ''}
-													onChange={setAlias}
-												/>
+												{email || ''}
```
silkederudder-shd, line 714: And here ;)

silkederudder-shd (changes requested):

src/settings/components/Profile.tsx
bertyhell: agreed, but lets use elvis operators instead of lodash get function since it provides better type safety
```
	const firstName = user?.first_name || '';
	const lastName = user?.last_name || '';
	const email = user?.mail || '';
```

src/settings/components/Profile.tsx
```diff
@@ -681,43 +682,34 @@ const Profile: FunctionComponent<
...
+										>
+											{firstName}
+										</FormGroup>
+										<Button
+											type="secondary"
+											label={t(
+												'settings/components/account___wijzig-gegevens'
+											)}
+											onClick={handleEditInfo}
 										/>
```
bertyhell, line 698: meemoo wants to use real links as much as possible. So if the url is known before hand, and you're trying to navigate to a page, use a Link element around the button instead of the onclick handler:
```suggestion
										<a href={generateEditProfileInfoLink()}>
											<Button
												type="secondary"
												label={t(
													'settings/components/account___wijzig-gegevens'
												)}
											/>
										</a>
```
That way, users can right click and select "open in new window". Which is something meemoo explicitly asked us to change in the past.

silkederudder-shd (approved):

bertyhell (approved):

PR #1680, opened 2023-01-17
fix(admin-core): move routes to admin-core config + make client boss

bertyhell:
Move the building of routes that are needed for the admin-core to the client
and pass the full to admin-core in the config object.

Requires admin-core PR: 
https://github.com/viaacode/react-admin-core-module/pull/88

also switched some content-page-labels related code to use the admin-core services instead (less code duplication) 
saves about 2500 lines of code


src/admin/content-page/views/ContentPageEditPage.tsx
```diff
@@ -1,29 +1,25 @@
...
 import { withAdminCoreConfig } from '../../shared/hoc/with-admin-core-config';
+import { CONTENT_PAGE_PATH } from '../content-page.consts';
 
 const ContentPageDetailPage: FunctionComponent<DefaultSecureRouteProps<ContentPageDetailProps>> = ({
 	history,
 	match,
 }) => {
 	const { id } = match.params;
-
+	console.log('match: ', JSON.stringify(match));
```
WardVerc1, line 16: Remove console.log?

src/dynamic-route-resolver/dynamic-route-resolver.const.ts
```diff
@@ -34,6 +34,8 @@ export const GET_REDIRECTS: () => { [avo1Path: string]: string } = () => ({
 
 	'/klaar.json': `${getEnv('PROXY_URL')}/klaar/klaar.json`,
 	'/projecten/klaar': '/projecten/uitgeklaard',
+	'/beheer': '/admin',
+	// '/beheer/.*': '/admin/.*',
```
WardVerc1, line 38: Remove comment?

silkederudder-shd (approved):

WardVerc1 (approved):

PR #1693, opened 2023-01-30
fix(avo-2355): delete ui that was moved to admin-core

bertyhell:
https://meemoo.atlassian.net/browse/AVO-2355

* Delete ui that was moved to the admin-core
* fix some routes that had a trailing }
* Add some translations
* extract translations => this also deletes the translations from the deleted admin-core ui screens


tourlbr (approved):

src/shared/components/MultiUserSelectDropdown/MultiUserSelectDropdown.tsx
```diff
@@ -62,12 +56,12 @@ export const MultiUserSelectDropdown: FunctionComponent<MultiUserSelectDropdownP
...
 		if (values.length) {
 			UserService.getNamesByProfileIds(values)
-				.then((users: Avo.User.User[]) => {
+				.then((users: Partial<CommonUser>[]) => {
 					setSelectedProfiles(
 						users.map(
 							(user): PickerItem => ({
 								label: `${get(user, 'full_name')} (${get(user, 'mail')})`,
-								value: user.profile?.id || '',
+								value: user.profileId || '',
```
silkederudder-shd, line 64: Maybe try use the same way of working throughout the project/file. Line 63 uses lodash to get a. value so maybe change that one to `${user.full_name} ${user.mail}` if you want to get rid of lodash.

silkederudder-shd (changes requested):

PR #1695, opened 2023-01-31
fix(avo-2365): fix admin layout style leakes

bertyhell:
https://meemoo.atlassian.net/browse/AVO-2365

before:
image

after:
image

### DIFF
--- a/src/admin/shared/layouts/AdminLayout/AdminLayout.scss
+++ b/src/admin/shared/layouts/AdminLayout/AdminLayout.scss
@@ -18,10 +18,10 @@
 
 	.m-admin-layout-content {
 		overflow-y: auto;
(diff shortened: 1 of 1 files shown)


tourlbr (approved):

src/admin/shared/layouts/AdminLayout/AdminLayout.scss
```diff
@@ -24,56 +24,56 @@
...
-			background-color: #eee;
+	.c-table.c-table_detail-page {
+		tr {
+			&:nth-child(even) {
+				background-color: #eee;
+			}
+
+			th,
+			td {
+				padding: 10px 10px;
```
WardVerc1, line 41: Are there variables to use? If not, use rem instead of px?

PR #1743, opened 2023-05-17
feat(avo-2597): Share component

SHDBramVandenbussche:

image


WardVerc1 (changes requested):
image
image

I get these weird fly out's

src/shared/components/ShareDropdown/ShareDropdown.tsx
```diff
@@ -0,0 +1,89 @@
...
+				label: tText('shared/components/share-dropdown/share-dropdown___collegas'),
+				icon: IconName.userTeacher,
+			},
+			{
+				id: 'pupils',
+				label: tText('shared/components/share-dropdown/share-dropdown___leerlingen'),
+				icon: IconName.userStudent,
+			},
+		],
+		'colleagues'
```
bertyhell, line 42: maybe change this to 'pupils' so we don't have to come back to this in the other task

src/shared/components/ShareDropdown/ShareDropdown.tsx
```diff
@@ -0,0 +1,89 @@
...
+
+	const handleShareButtonClicked = () => {
+		setIsShareDropdownOpen(!isShareDropdownOpen);
+	};
+
+	return (
+		<Dropdown
+			isOpen={isShareDropdownOpen}
+			onClose={() => setIsShareDropdownOpen(false)}
+			className="c-share-dropdown"
```
bertyhell, line 53: You can add an option to show the dropdown right aligned to the button position:
```
placement="bottom-end"
```

image


so it matches the design:
https://www.figma.com/file/xWFZmBGjGetTJOlIqnaeEY/Delen-%26-Samenwerken-(MASTER-%E2%80%94-Internal-use-only)?type=design&node-id=1037-7792&t=e4WArnelsp31sfCr-0

src/shared/components/ShareDropdown/ShareDropdown.tsx
```diff
@@ -0,0 +1,89 @@
...
+					disabled={false}
+				/>
+			</DropdownButton>
+
+			<DropdownContent>
+				<div className="c-share-dropdown__container">
+					<Tabs tabs={tabs} onClick={(id) => setActiveTab(id)} />
+
+					<div className="c-share-dropdown__content">
+						{tab === 'colleagues' ? (
```
bertyhell, line 72: make an enum for colleagues and pupils tabs, so we can reuse those values and know where they are used easily

src/shared/components/ShareWithColleagues/ShareWithColleagues.helpers.tsx
```diff
@@ -0,0 +1,12 @@
+import { flatten, groupBy } from 'lodash';
```
bertyhell, line 1: always import from 'lodash-es' since it supports tree shaking and makes the final javascript file bundle smaller

src/shared/components/ShareWithColleagues/ShareWithColleagues.helpers.tsx
```diff
@@ -0,0 +1,12 @@
...
+
+import { ShareUserInfo, ShareUserInfoRights } from './ShareWithColleagues.types';
+
+export const sortShareUsers = (users: ShareUserInfo[]) => {
+	const groupedUsers = groupBy(users, 'rights');
+	return flatten([groupedUsers.EIGENAAR, groupedUsers.BIJDRAGER, groupedUsers.KIJKER]);
```
bertyhell, line 7: You can also destruct the arrays into the parent array.
```suggestion
	return [
		...groupedUsers[ShareUserInfoRights.OWNER],
		...groupedUsers[ShareUserInfoRights.CONTRIBUTOR],
		...groupedUsers[ShareUserInfoRights.VIEWER]
	];
```

src/shared/components/ShareWithColleagues/ShareWithColleagues.helpers.tsx
```diff
@@ -0,0 +1,12 @@
...
+};
+
+export const shareUserRightToString = (right: ShareUserInfoRights) => {
+	return (right as unknown as string).toLocaleLowerCase();
```
bertyhell, line 11:
```suggestion
	return right.toLowerCase();
```

src/shared/components/ShareWithColleagues/ShareWithColleagues.tsx
```diff
@@ -0,0 +1,225 @@
...
+	Button,
+	Dropdown,
+	DropdownButton,
+	DropdownContent,
+	Icon,
+	IconName,
+	MenuContent,
+	TextInput,
+} from '@viaa/avo2-components';
+import { isEmpty, isNil, truncate } from 'lodash';
```
bertyhell, line 12:
```suggestion
import { isEmpty, isNil, truncate } from 'lodash-es';
```

src/shared/components/ShareWithColleagues/ShareWithColleagues.tsx
```diff
@@ -0,0 +1,225 @@
...
+		rights: undefined,
+	});
+	const [error, setError] = useState<string | null>(null);
+
+	const handleRightsButtonClicked = () => {
+		setIsRightsDropdownOpen(!isRightsDropdownOpen);
+	};
+
+	const handleAddNewUser = () => {
+		if (/^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]+$/.test(newShareUser.email as string)) {
```
bertyhell, line 50: if email is undefined this code will crash
it's better to just default to empty string, then the regex test fill execute normally and return false

We might want to extract this as a helper in the shared folder since it is also used here. Something like: isValidEmail(): boolean
```
		if (!email) {
			errors.push(`Email ${requiredError}`);
		} else if (!/^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]+$/.test(email)) {
			errors.push(
				tText(
					'authentication/views/registration-flow/r-4-manual-registration___email-is-geen-geldig-email-adres'
				)
			);
		}
```

```suggestion
		if (/^[a-zA-Z0-9._%+-]+@[a-zA-Z0-9.-]+\.[a-zA-Z]+$/.test(newShareUser.email || '')) {
```

src/shared/components/ShareWithColleagues/ShareWithColleagues.tsx
```diff
@@ -0,0 +1,225 @@
...
+			onAddNewUser(newShareUser);
```
bertyhell, line 51: We also want to reset the local state, since the newly added user will be added to the users list:
```suggestion
			await onAddNewUser(newShareUser);
			setNewShareUser({ email: undefined, rights: undefined });
```

src/shared/components/ShareWithColleagues/ShareWithColleagues.tsx
```diff
@@ -0,0 +1,225 @@
...
+							<div className="c-colleague-info-row__info">
+								{(user.firstName || user.lastName) && (
+									<p>{`${user.firstName} ${user.lastName}`}</p>
+								)}
+
+								<p className="c-colleague-info-row__info__email">
+									{truncate(user.email, {
+										length: 32,
+										separator: ' ',
+									})}
```
bertyhell, line 103: wouldn't this be better truncated with css: 
https://developer.mozilla.org/en-US/docs/Web/CSS/text-overflow

src/shared/components/ShareWithColleagues/ShareWithColleagues.tsx
```diff
@@ -0,0 +1,225 @@
...
+			...newShareUser,
+			...value,
+		});
+	};
+
+	const renderColleaguesInfoList = () => {
+		if (users.length > 0) {
+			return (
+				<ul className="c-colleagues-info-list">
+					{sortShareUsers(users).map((user, index) => (
```
bertyhell, line 81: the user variable name is a naming conflict with the user parameter of the component
this doesn't break anything, but it is better to avoid it since it can cause confusion
```suggestion
					{sortShareUsers(users).map((contributorUser, index) => (
```

src/shared/components/ShareWithColleagues/ShareWithColleagues.tsx
```diff
@@ -0,0 +1,225 @@
...
+											<Icon name={IconName.edit2} />
+										</button>
+									)}
+							</div>
+
+							<div className="c-colleague-info-row__action">
+								{((currentUser.rights === ShareUserInfoRights.OWNER &&
+									currentUser.email !== user.email) ||
+									(currentUser.rights !== ShareUserInfoRights.OWNER &&
+										currentUser.email === user.email)) && (
```
bertyhell, line 130: Add a comment or create a variable at the start of the map function:
```
// Only the owner can delete any user
// And contributor and viewers can only delete themselves
const canDelete = 
	(currentUser.rights === ShareUserInfoRights.OWNER && currentUser.email !== user.email) ||
	(currentUser.rights !== ShareUserInfoRights.OWNER && currentUser.email === user.email)
```

src/shared/components/ShareWithColleagues/ShareWithColleagues.tsx
```diff
@@ -0,0 +1,225 @@
...
+						</Dropdown>
+
+						<Button
+							icon={IconName.add}
+							label={tText(
+								'shared/components/share-with-colleagues/share-with-colleagues___voeg-toe'
+							)}
+							className="c-add-colleague__button"
+							onClick={handleAddNewUser}
+							disabled={isEmpty(newShareUser.email) || isNil(newShareUser.rights)}
```
bertyhell, line 212:
```suggestion
							disabled={!isValidEmail(newShareUser.email) || !newShareUser.rights}
```

bertyhell (changes requested):

bertyhell (approved):

src/shared/components/ShareDropdown/ShareDropdown.tsx
SHDBramVandenbussche: I have fixed  this in another branch

PR #1761, opened 2023-06-08
feat(avo-2511): change collection to new overview table

WardVerc1:
Ticket: https://meemoo.atlassian.net/browse/AVO-2511

### DIFF
--- a/src/collection/queries/get-collections-by-owner.graphql
+++ b/src/collection/queries/get-collections-by-owner.graphql
@@ -3,10 +3,10 @@ query getCollectionsByOwner(
 	$type_id: Int
 	$offset: Int = 0
 	$limit: Int
-	$order: [app_collections_order_by!] = { updated_at: desc }
-	$where: [app_collections_bool_exp!] = []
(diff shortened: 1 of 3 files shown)


SHDBramVandenbussche (approved):

src/collection/queries/get-collections-by-owner.graphql
```diff
@@ -68,5 +68,25 @@ query getCollectionsByOwner(
 				}
 			}
 		}
+		share_type
+		contributors {
+			enum_right_type {
+				value
+			}
+			profile {
+				usersByuserId {
```
bertyhell, line 77: OPTIONAL: this relationship is named a bit funky in the database
we can add an alias to use user instead in the query
This allows us to use `contributor.profile.user` instead of `contributor.profile.usersByuserId` in the code

and that means you might be able to use Avo.User.Profile to type these fields instead of having to use custom types.
```suggestion
				user: usersByuserId {
```

bertyhell (approved):

PR #1762, opened 2023-06-08
feat(avo-2622): added share_type column to collections overview

WardVerc1:
image
image
image

Ticket: https://meemoo.atlassian.net/browse/AVO-2622
Proxy: https://github.com/viaacode/avo2-proxy/pull/633


src/collection/components/CollectionOrBundleOverview.tsx
```diff
@@ -501,9 +504,48 @@ const CollectionOrBundleOverview: FunctionComponent<CollectionOrBundleOverviewPr
...
 	};
 
-	const renderCell = (collection: Avo.Collection.Collection, colKey: string) => {
+	const renderCell = (collection: Collection, colKey: string) => {
 		const { id } = collection;
 
+		const sharedWithNames = collection.contributors.map((contributor) => {
+			if (contributor.profile?.organisation?.name) {
+				return (
+					contributor.profile?.usersByuserId?.full_name +
```
SHDBramVandenbussche, line 513: Might be better to use template literals here ``${fullName} (${org})  ``

SHDBramVandenbussche (approved):

src/collection/collection.types.ts
```diff
@@ -10,7 +10,7 @@ import {
 export type Collection = (
 	| GetPublicCollectionsByIdQuery
 	| GetPublicCollectionsByTitleQuery
-)['app_collections'][0];
+)['app_collections_overview'][0];
```
bertyhell, line 13: why do these queries use the overview table? Do we need to the share_type somewhere for these?

src/assignment/modals/AddCollectionModal.tsx
```diff
@@ -137,7 +137,7 @@ const AddCollectionModal: FunctionComponent<AddCollectionModalProps> = ({
 			const columnDataType: TableColumnDataType = (column?.dataType ||
 				TableColumnDataType.string) as TableColumnDataType;
 
-			let collections: Avo.Collection.Collection[];
+			let collections: Collection[];
 			if (activeView === AddCollectionTab.myCollections) {
 				collections = await CollectionService.fetchCollectionsByOwner(
```
bertyhell, line 142: is this function name still acurate? Shouldn't this be something like:
fetchCollectionsByOwnerOrContributorProfileId?

src/collection/collection.service.ts
```diff
@@ -1279,7 +1279,7 @@ export class CollectionService {
 		order: Record<string, 'asc' | 'desc'> | Record<string, 'asc' | 'desc'>[],
 		contentTypeId: ContentTypeNumber.collection | ContentTypeNumber.bundle,
 		filterString: string | undefined
-	): Promise<Avo.Collection.Collection[]> {
+	): Promise<Collection[]> {
```
bertyhell, line 1282: OPTIONAL: it might be better to add the share_type?: ... to the Avo.Collection.Collection type, so we can keep using a single type?
only if this isn't too much work to change.

src/collection/components/CollectionOrBundleOverview.tsx
```diff
@@ -501,9 +504,44 @@ const CollectionOrBundleOverview: FunctionComponent<CollectionOrBundleOverviewPr
...
+				return `${fullName} `;
+			}
+		});
+
+		const shareTypeTitle =
+			collection.share_type === CollectionShareType.GEDEELD_MET_MIJ
+				? tText('collection/views/collection-overview___gedeeld-met-mij')
+				: collection.share_type === CollectionShareType.GEDEELD_MET_ANDERE
+				? tText('collection/views/collection-overview___gedeeld-met-anderen')
+				: tText('collection/views/collection-overview___mijn-collectie');
```
bertyhell, line 526: OPTIONAL: i sometimes use this notation instead of a double tertiary operator since it is slightly more readable. The same applies to the 2 consts below
```suggestion
		const shareTypeTitle = {
			[CollectionShareType.GEDEELD_MET_MIJ]: tText('collection/views/collection-overview___gedeeld-met-mij'),
			[CollectionShareType.GEDEELD_MET_ANDERE]: tText('collection/views/collection-overview___gedeeld-met-anderen'),
			[CollectionShareType.NIET_GEDEELD]: tText('collection/views/collection-overview___mijn-collectie'),
		}[collection.share_type];
```

bertyhell (approved):

src/collection/collection.types.ts
WardVerc1: Yes is needed, I got typing issues with Collection

src/collection/collection.service.ts
WardVerc1: I tried to keep it consistent/aligned with assignments

PR #1766, opened 2023-06-12
feat(avo-2618): added share type filter to assignment overview page

WardVerc1:
image
image
image
Ticket: https://meemoo.atlassian.net/browse/AVO-2618


src/assignment/assignment.service.ts
```diff
@@ -175,6 +176,11 @@ export class AssignmentService {
 					labels: { assignment_label_id: { _in: classIds } },
 				});
 			}
+			if (shareTypeIds && shareTypeIds.length) {
```
bertyhell, line 179:
```suggestion
			if (shareTypeIds?.length) {
```

src/assignment/views/AssignmentOverview.tsx
```diff
@@ -644,6 +652,26 @@ const AssignmentOverview: FunctionComponent<AssignmentOverviewProps> = ({
...
+				{
+					// add GEDEELD MET ANDEREN to this option
+					label: tText('assignment/views/assignment-overview___mijn-opdrachten'),
+					id: AssignmentShareType.NIET_GEDEELD,
+					checked: [...(query.selectedShareTypeLabelIds || [])].includes(
+						AssignmentShareType.NIET_GEDEELD
+					),
+				},
+			]);
+		}
```
bertyhell, line 673: the currently contains both CLASS and LABEL because those labels are stored in the same database field

So you better put the share type labels in a separate function since that is a separate database field and has nothing to do with CLASS/LEBEL assignment labels

```typescript
	getShareTypeLabels = (): CheckboxOption[] => {
		return compact([
			{
				label: tText('assignment/views/assignment-overview___gedeeld-met-mij'),
				id: AssignmentShareType.GEDEELD_MET_MIJ,
				checked: [...(query.selectedShareTypeLabelIds || [])].includes(
					AssignmentShareType.GEDEELD_MET_MIJ
				),
			},
			{
				// add GEDEELD MET ANDEREN to this option
				label: tText('assignment/views/assignment-overview___mijn-opdrachten'),
				id: AssignmentShareType.NIET_GEDEELD,
				checked: [...(query.selectedShareTypeLabelIds || [])].includes(
					AssignmentShareType.NIET_GEDEELD
				),
			},
		]);
	}
```

bertyhell (changes requested):

bertyhell (approved):

PR #1768, opened 2023-06-13
fix(update-assignment-function): Replace return of map with lom lom id

SHDBramVandenbussche:


src/assignment/assignment.service.ts
```diff
@@ -410,10 +410,12 @@ export class AssignmentService {
 				original.blocks || [],
 				update.blocks || []
 			);
+			console.log(update);
 
 			await AssignmentService.deleteAssignmentLomLinks(original.id);
 
-			const loms = map(update.loms, 'lom_id');
+			const loms = update.loms.map((lom) => lom.lom.id);
+			console.log(loms);
```
WardVerc1, line 418: Remove console.log

WardVerc1 (approved):

bertyhell (approved):

PR #1770, opened 2023-06-13
feat(avo-2669): add max length for labels

SHDBramVandenbussche:
[AVO-2669](https://meemoo.atlassian.net/browse/AVO-2669)

Before:
image

After:
image

[AVO-2669]: https://meemoo.atlassian.net/browse/AVO-2669?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/assignment/components/modals/ManageAssignmentLabels.tsx
```diff
@@ -134,12 +136,29 @@ const ManageAssignmentLabels: FunctionComponent<ManageAssignmentLabelsProps> = (
 		setAssignmentLabels([...assignmentLabels.filter((labelObj) => labelObj.id !== id)]);
 	};
 
+	const labelsExceedMaxLenght = (labels: Assignment_Label_v2[]) => {
```
bertyhell, line 139:
```suggestion
	const labelsExceedMaxLength = (labels: Assignment_Label_v2[]) => {
```

src/assignment/components/modals/ManageAssignmentLabels.tsx
```diff
@@ -134,12 +136,29 @@ const ManageAssignmentLabels: FunctionComponent<ManageAssignmentLabelsProps> = (
...
+		let isValid = true;
+
+		labels.forEach((label) => {
+			const exceedsMaxLength = (label.label?.length || 0) > MAX_LABEL_LENGTH;
+			if (exceedsMaxLength) {
+				isValid = false;
+			}
+		});
+
+		return isValid;
```
bertyhell, line 149: find is more efficient than a foreach since it will stop after finding the first occurance
```suggestion
		return !labels.find(label => (label.label?.length || 0) > MAX_LABEL_LENGTH);
```

src/assignment/components/modals/ManageAssignmentLabels.tsx
```diff
@@ -134,12 +136,29 @@ const ManageAssignmentLabels: FunctionComponent<ManageAssignmentLabelsProps> = (
...
+
 	const handleSaveLabels = async () => {
 		try {
 			setIsProcessing(true);
 			const initialAssignmentLabelIds = initialAssignmentLabels.map((l) => l.id);
 			const updatedAssignmentLabelIds = assignmentLabels.map((l) => l.id);
 
+			if (!labelsExceedMaxLenght(assignmentLabels)) {
+				throw new CustomError('One or more labels is longer then max length', 'invalid');
+			}
```
bertyhell, line 160: preferably we don't use exceptions for handing logic we expect to happen. Since exceptions are quite heavy to throw en catch.
So if there is an easy alternative, we prefer a simple if + return;

then you can also simplify the catch statement at the end
```suggestion
			if (!labelsExceedMaxLenght(assignmentLabels)) {
				ToastService.danger(
					tHtml('Een of meerdere labels is langer dan {{maxLength}} karakters', {
						maxLength: MAX_LABEL_LENGTH,
					})
				);
				setIsProcessing(false);
				return;
			}
```

PR #1771, opened 2023-06-13
feat(avo-2620): added share type filter on collection overview page

WardVerc1:
image
image
Ticket: https://meemoo.atlassian.net/browse/AVO-2620

### DIFF
--- a/src/collection/collection.service.ts
+++ b/src/collection/collection.service.ts
@@ -1288,13 +1288,14 @@ export class CollectionService {
 		}
 	}
 
-	static async fetchCollectionsByOwner(
+	static async fetchCollectionsByOwnerOrContributorProfileId(
 		user: Avo.User.User,
 		offset: number,
 		limit: number | null,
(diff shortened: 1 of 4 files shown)


src/collection/collection.service.ts
```diff
@@ -1288,13 +1288,14 @@ export class CollectionService {
...
-	static async fetchCollectionsByOwner(
+	static async fetchCollectionsByOwnerOrContributorProfileId(
 		user: Avo.User.User,
 		offset: number,
 		limit: number | null,
 		order: Record<string, 'asc' | 'desc'> | Record<string, 'asc' | 'desc'>[],
 		contentTypeId: ContentTypeNumber.collection | ContentTypeNumber.bundle,
-		filterString: string | undefined
+		filterString: string | undefined,
+		shareTypeIds?: string[] | undefined
```
SHDBramVandenbussche, line 1298: When you use optional parameters doesn't that make them of type ... | undefined  as default? I don't think you can just give line 1297 and 1298 a type of string / string[] and make them optional.

SHDBramVandenbussche (approved):

PR #1777, opened 2023-06-16
fix: general fixes

SHDBramVandenbussche:
- Delete loms and contributors keys from assignment object before updating, fix typing issue
- Add conditional render to publish en share button

### DIFF
--- a/src/assignment/components/AssignmentActions.tsx
+++ b/src/assignment/components/AssignmentActions.tsx
@@ -10,6 +10,7 @@ import type { Avo } from '@viaa/avo2-types';
 import classNames from 'classnames';
 import React, { FunctionComponent, useCallback, useEffect, useMemo, useState } from 'react';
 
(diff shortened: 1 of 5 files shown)


src/assignment/components/AssignmentActions.tsx
```diff
@@ -176,30 +179,36 @@ const AssignmentActions: FunctionComponent<AssignmentActionsProps> = ({
 		/>
 	);
 
-	const renderPublishButton = (config?: Partial<ButtonProps>) => (
-		<Button type="secondary" {...config} />
-	);
+	const renderPublishButton = (config?: Partial<ButtonProps>) => {
+		if (route !== APP_PATH.ASSIGNMENT_DETAIL.route) {
```
bertyhell, line 183: you're checking if the current page is the detail route
but this component isn't used on the detail page

is this a mistake?

bertyhell (approved):

PR #1800, opened 2023-06-30
feat(avo-2652): Add an attempt to request edit status and redirect in case of fail with toast message

SHDBramVandenbussche:
[AVO-2652](https://meemoo.atlassian.net/browse/AVO-2652)
[AVO-2661](https://meemoo.atlassian.net/browse/AVO-2661)

[AVO-2652]: https://meemoo.atlassian.net/browse/AVO-2652?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[AVO-2661]: https://meemoo.atlassian.net/browse/AVO-2661?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/assignment/views/AssignmentEdit.tsx
+++ b/src/assignment/views/AssignmentEdit.tsx
@@ -76,7 +76,7 @@ import AssignmentResponses from './AssignmentResponses';
 
 import './AssignmentEdit.scss';
 import './AssignmentPage.scss';
-import { buildLink } from '../../shared/helpers';
+import { buildLink, CustomError } from '../../shared/helpers';
 import { InActivityWarningModal } from '../../shared/components';
 
 interface AssignmentEditProps extends DefaultSecureRouteProps<{ id: string; tabId: string }> {
@@ -135,11 +135,15 @@ const AssignmentEdit: FunctionComponent<AssignmentEditProps> = ({
 		updateBlocksInAssignmentState
 	);
 
-	useEffect(() => {
+	const updateAssignmentEditorWithLoading = useCallback(async () => {
 		setAssigmentLoading(true);
-		updateAssignmentEditor();
+		await updateAssignmentEditor();
 		setAssigmentLoading(false);
-	}, []);
+	}, [setAssigmentLoading]);
+
+	useEffect(() => {
+		updateAssignmentEditorWithLoading();
+	}, [updateAssignmentEditorWithLoading]);
 
 	useEffect(() => {
 		const param = match.params.tabId;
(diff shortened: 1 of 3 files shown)


WardVerc1 (approved):

src/assignment/views/AssignmentEdit.tsx
```diff
@@ -135,6 +135,12 @@ const AssignmentEdit: FunctionComponent<AssignmentEditProps> = ({
 		updateBlocksInAssignmentState
 	);
 
+	useEffect(() => {
+		setAssigmentLoading(true);
+		updateAssignmentEditor();
+		setAssigmentLoading(false);
+	}, []);
```
bertyhell, line 142: if you don't wait for the updateAssignmentEditor then there is no point in setting the loading state, since it will happen instantly
you can await inside a useEffect, so my solution is:
```suggestion
	const updateAssignmentEditorWithLoading = useCallback(async () => {
		setAssigmentLoading(true);
		await updateAssignmentEditor();
		setAssigmentLoading(false);
	}, [setAssigmentLoading]);
	
	useEffect(() => {
		updateAssignmentEditorWithLoading();
	}, [updateAssignmentEditorWithLoading]);
```

or maybe set the loading state inside the updateAssignmentEditor function?

src/assignment/views/AssignmentEdit.tsx
```diff
@@ -734,11 +738,11 @@ const AssignmentEdit: FunctionComponent<AssignmentEditProps> = ({
...
-			!canEditAllAssignments ||
-			(assignment &&
-				!isUserAssignmentOwner(user, assignment) &&
-				!isUserAssignmentContributor(user, assignment) &&
-				!canEditAllAssignments)
+			// !canEditAllAssignments ||
+			assignment &&
+			!isUserAssignmentOwner(user, assignment) &&
+			!isUserAssignmentContributor(user, assignment) &&
+			!canEditAllAssignments
```
bertyhell, line 745: Assignment is always defined, because it is the form state
You want to check for an id to be sure it is an exisiting assignment that has been fetched from the database
```suggestion
			!canEditAllAssignments &&
			assignment?.id &&
			!isUserAssignmentOwner(user, assignment) &&
			!isUserAssignmentContributor(user, assignment)
```

src/collection/components/CollectionOrBundleEdit.tsx
```diff
@@ -195,6 +195,12 @@ const CollectionOrBundleEdit: FunctionComponent<
 		icon: IconName.alertTriangle,
 	} as LoadingInfo;
 
+	useEffect(() => {
+		setLoadingInfo({ state: 'loading' });
+		updateCollectionEditor();
+		setLoadingInfo({ state: 'loaded' });
+	}, []);
```
bertyhell, line 202: preferably we use a react-query wrapper for this, so we get the loading state from the library for free.
eg: https://github.com/viaacode/hetarchief-client/blob/master/src/modules/visits/hooks/update-visit.ts

bertyhell (changes requested):

src/assignment/views/AssignmentEdit.tsx
SHDBramVandenbussche: assignment is possibly undefined somehow and typescript complains about it if i remove the check
image

bertyhell (approved):

PR #1802, opened 2023-06-30
feat(avo-2670): add avatar and author to collection, bundle and assignment searchresults

SHDBramVandenbussche:
[AVO-2670](https://meemoo.atlassian.net/browse/AVO-2670)

[Avo2-typings PR](https://github.com/viaacode/avo2-typings/pull/127)

Before:
image

After: 
image

[AVO-2670]: https://meemoo.atlassian.net/browse/AVO-2670?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/search/components/SearchResultItem.tsx
```diff
@@ -64,14 +65,42 @@ const SearchResultItem: FunctionComponent<SearchResultItemProps> = ({
 			/>
 		);
 	};
+	const checkIfResultIsNotAnItem = (result: Avo.Search.ResultItem): boolean => {
+		return (
+			result.administrative_type === 'collectie' ||
+			result.administrative_type === 'bundel' ||
+			result.administrative_type === 'opdracht'
```
WardVerc1, line 72: Can this be written shorter so you don't have to write `result.administrative_type` 3x?
Something like `result.administrative_type === ('collectie' || 'bundel' || 'opdracht')` ?

WardVerc1 (approved):

src/search/components/SearchResultItem.tsx
```diff
@@ -64,14 +65,38 @@ const SearchResultItem: FunctionComponent<SearchResultItemProps> = ({
 			/>
 		);
 	};
+	const checkIfResultIsNotAnItem = (result: Avo.Search.ResultItem): boolean => {
+		return result.administrative_type === ('collectie' || 'bundel' || 'opdracht');
```
bertyhell, line 69: this doesn't work. you should think about this from an order of execution perspective. The thing between the parenteses is evaluated first.
so: 
* 'collectie' || 'bundel' => truthy || truthy => true
* true || 'opdracht' => true
* result.administrative_type === true => false

This will always be false

the correct way:
```suggestion
		return ['collectie', 'bundel', 'opdracht'].includes(result.administrative_type);
```

src/search/components/SearchResultItem.tsx
```diff
@@ -64,14 +65,38 @@ const SearchResultItem: FunctionComponent<SearchResultItemProps> = ({
...
+	};
+
+	const renderAuthorOrOrganisation = (result: Avo.Search.ResultItem) => {
+		if (checkIfResultIsNotAnItem(result)) {
```
bertyhell, line 73: Prefer to avoid negatives in function names, since they are harder to reason about.
```suggestion
		if (!isItem(result)) {
```

and flip the inner boolean inside the function or change the logic to check for items instead of not items

src/search/components/SearchResultItem.tsx
```diff
@@ -64,14 +65,38 @@ const SearchResultItem: FunctionComponent<SearchResultItemProps> = ({
...
+			const name = `${result.owner?.firstname} ${result.owner?.lastname}`;
+			const initials = `${result.owner?.firstname?.[0]}${result.owner?.lastname?.[0]}`;
```
bertyhell, line 75: When you use the elvis operator: ?.
you need to provide a default otherwise you'll be working with undefined
```suggestion
			const name = `${result.owner?.firstname || 'Naamloos'} ${result.owner?.lastname || ''}`;
			const initials = `${result.owner?.firstname?.[0] || ''}${result.owner?.lastname?.[0] || ''}`;
```

src/search/components/SearchResultItem.tsx
```diff
@@ -64,14 +65,38 @@ const SearchResultItem: FunctionComponent<SearchResultItemProps> = ({
...
+		}
+
+		if (result.original_cp) {
+			return renderSearchLink(
+				result.original_cp,
+				{ filters: { provider: [result.original_cp] } },
+				'c-body-2'
+			);
+		}
+	};
```
bertyhell, line 95: What happens if both if statements are false
We should return something, even if it is just: return null;

src/search/components/SearchResultItem.tsx
```diff
@@ -100,15 +125,15 @@ const SearchResultItem: FunctionComponent<SearchResultItemProps> = ({
...
-					</SearchResultSubtitle>
-				)}
+				<SearchResultSubtitle>
+					{result.original_cp
+						? renderSearchLink(
+								result.original_cp,
+								{ filters: { provider: [result.original_cp] } },
+								'c-body-2'
+						  )
+						: renderAuthorOrOrganisation(result)}
```
bertyhell, line 135: this seems to be a bit double with the result.original_cp check in renderAuthorOrOrganisation
maybe we need to split that part off into a function: renderCpLink?

bertyhell (changes requested):

bertyhell (approved):

PR #1805, opened 2023-07-03
fix(avo-2597): fix width role dropdown button in share with colleagues component

SHDBramVandenbussche:

Before: 
image

After:
image


WardVerc1 (approved):

src/shared/components/ShareWithColleagues/ShareWithColleagues.scss
```diff
@@ -19,30 +19,36 @@ $borderLight: 1px solid $color-gray-100;
...
-		text-transform: capitalize;
+	&__select.c-button {
+		.c-button__content {
+			justify-content: space-between;
+		}
 
-		&:hover,
-		:active,
-		:active:focus {
+		&.c-button--primary {
```
bertyhell, line 27: we should also fix the hover issue:
image

image

The correct way of doing this is to use Button type="secondary" or maybe tertiary
and not override the styling of the primary button style

bertyhell (changes requested):

bertyhell (approved):

PR #1806, opened 2023-07-03
fix(avo-2620): Move render header function in orde to display filter in every situation

SHDBramVandenbussche:
AVO-[2620](https://meemoo.atlassian.net/browse/AVO-2620)

Before: 
image

After:
image


WardVerc1 (approved):

WardVerc1:
Weird that we have to show this when there is nothing to show/sort, but if it's what they want 🤷‍♂️

SHDBramVandenbussche:
> Weird that we have to show this when there is nothing to show/sort, but if it's what they want 🤷‍♂️

We have to be able to clear the filter when one of them returns nothing

WardVerc1:
Ah makes sense 👌

bertyhell (approved):

PR #1815, opened 2023-07-07
fix(avo-2608): owner and contributors spacing

SHDBramVandenbussche:
[AVO-2608](https://meemoo.atlassian.net/browse/AVO-2608)

[AVO-2608]: https://meemoo.atlassian.net/browse/AVO-2608?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/shared/components/HeaderOwnerAndContributors/HeaderOwnerAndContributors.tsx
+++ b/src/shared/components/HeaderOwnerAndContributors/HeaderOwnerAndContributors.tsx
@@ -10,7 +10,7 @@ import { Avo } from '@viaa/avo2-types';
 import React, { FC, ReactNode } from 'react';
 
 import { getFullName } from '../../helpers';
-import { tHtml } from '../../helpers/translate';
+import { tHtml, tText } from '../../helpers/translate';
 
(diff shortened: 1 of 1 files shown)


bertyhell (approved):

src/shared/components/HeaderOwnerAndContributors/HeaderOwnerAndContributors.tsx
```diff
@@ -28,15 +28,21 @@ export const HeaderOwnerAndContributors: FC<HeaderOwnerAndContributorsProps> = (
 
 	const renderContributors = (): ReactNode => {
 		if (contributors?.length) {
+			const couplingWord = tText('en');
```
bertyhell, line 31: fyi, you don't need to create a variable for this neccessarily since the key that will be generated for identical translations in identical files will be the same, so only one translation entry will be made in nl.json

But in this case the variable will save some extra bytes :D

bertyhell (approved):

PR #1819, opened 2023-07-07
fix(2597): adjust styles of share with colleague component

SHDBramVandenbussche:

Before:
image

image

After:
image

image

### DIFF
--- a/src/shared/components/ShareWithColleagues/ShareWithColleagues.scss
+++ b/src/shared/components/ShareWithColleagues/ShareWithColleagues.scss
@@ -4,6 +4,7 @@ $borderLight: 1px solid $color-gray-100;
 
 .c-add-colleague {
 	display: flex;
+	margin-bottom: 2rem;
 
 	.c-button.c-button--primary {
 		border-radius: 0;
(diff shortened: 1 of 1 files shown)


src/shared/components/ShareWithColleagues/ShareWithColleagues.scss
```diff
@@ -57,7 +58,7 @@ $borderLight: 1px solid $color-gray-100;
 		border: $borderLight;
 		border-radius: 0 $c-input-border-radius $c-input-border-radius 0;
 		background-color: $color-gray-50;
-		margin-left: 0;
+		margin-left: 0 !important;
```
bertyhell, line 61: usually we try to avoid !important, since it can be hard to overcome if this selector applies to some other part of the code that you don't want it to apply to

a better solution is to make the current selector more specific

the current selector is:
```
.c-add-colleague__select.c-button.c-button--primary
```

Which contains 2 classes, so it has a specificity of 0 - 2 - 0 ( https://developer.mozilla.org/en-US/docs/Web/CSS/Specificity )

While the other style that adds a margin left to the button also has a specificity of 0 - 2 - 0, so the order of the css in the final css file is used to determine which should be applied


We can fix this by making the specificty of the current rule 0 - 3 - 0 by adding another class on line 5
```
.c-dropdown__content-open .c-add-colleague {
```

bertyhell (approved):

PR #1822, opened 2023-07-10
feat(avo-2645): rework collection buttons visibility and add share buttons

bertyhell:
https://meemoo.atlassian.net/browse/AVO-2645


SHDBramVandenbussche (approved):

src/styles/components/_dropdown.scss
```diff
@@ -0,0 +1,4 @@
+.c-dropdown__content-open {
+	// Move the open dropdown above the tabs component
+	z-index: 11;
```
WardVerc1, line 3: We have a function `get-z-layer` to get a z-index. In `_variables.scss` you can see the options.

src/assignment/helpers/assignment-share-with-collegue-handlers.ts
```diff
@@ -0,0 +1,110 @@
...
+export async function onEditContributor(
+	contributor: ContributorInfo,
+	newRights: ShareRightsType,
+	shareWithPupilsProps: ShareWithPupilsProps,
+	fetchContributors: () => void,
+	refetchAssignment: () => void
+) {
+	try {
+		if (shareWithPupilsProps && refetchAssignment) {
+			if (newRights === 'OWNER') {
```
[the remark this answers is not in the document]
bertyhell: will fix in: https://github.com/viaacode/avo2-client/pull/1824

src/styles/components/_dropdown.scss
bertyhell: i don't see that function. Maybe you're thinking of hetarchief instead of avo?

PR #1855, opened 2023-07-20
fix(avo-2792): replace existing check with permission service check and refactor

SHDBramVandenbussche:
[AVO-2792](https://meemoo.atlassian.net/browse/AVO-2792)

[AVO-2792]: https://meemoo.atlassian.net/browse/AVO-2792?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/assignment/views/AssignmentEdit.tsx
+++ b/src/assignment/views/AssignmentEdit.tsx
@@ -241,11 +241,7 @@ const AssignmentEdit: FunctionComponent<AssignmentEditProps> = ({
 
 			const checkedPermissions = await PermissionService.checkPermissions(
 				{
-					canEdit: [
-						{
-							name: PermissionName.EDIT_OWN_ASSIGNMENTS,
-							obj: assignmentId,
-						},
+					canEditAllAssignments: [
 						{
 							name: PermissionName.EDIT_ANY_ASSIGNMENTS,
 						},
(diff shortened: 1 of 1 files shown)


src/assignment/views/AssignmentEdit.tsx
```diff
@@ -238,34 +237,15 @@ const AssignmentEdit: FunctionComponent<AssignmentEditProps> = ({
...
 			const hasPupilBlocks = await AssignmentService.hasPupilCollectionBlocks(id);
 
 			const checkedPermissions = await PermissionService.checkPermissions(
 				{
-					canEditAllAssignments: [
+					canEdit: [
+						{
+							name: PermissionName.EDIT_OWN_ASSIGNMENTS,
+							obj: assignmentId,
+						},
```
bertyhell, line 248: this adds back an additional call to the backend to fetch the assignment
but we already fetched it, and the backend checked that we can view the assignment

so the old code was more efficient

src/assignment/views/AssignmentEdit.tsx
```diff
@@ -283,6 +263,18 @@ const AssignmentEdit: FunctionComponent<AssignmentEditProps> = ({
 				user
 			);
 
+			if (!checkedPermissions.canEdit) {
```
bertyhell, line 266: you can store this in a computed variable:
```
const canEditAssignment = assignment?.id &&
			!isUserAssignmentOwner(user, assignment) &&
			!isUserAssignmentContributor(user, assignment) &&
			!permissions.canEditAllAssignments;
```

bertyhell (changes requested):

bertyhell (approved):

PR #1900, opened 2023-11-27
fix(avo-2934): do not release edit status after delete

WardVerc1:
image

Ticket: https://meemoo.atlassian.net/browse/AVO-2934

### DIFF
--- a/src/assignment/views/AssignmentEdit.tsx
+++ b/src/assignment/views/AssignmentEdit.tsx
@@ -891,7 +891,9 @@ const AssignmentEdit: FunctionComponent<AssignmentEditProps & UserProps> = ({
 
 					<InActivityWarningModal
 						onActivity={updateAssignmentEditor}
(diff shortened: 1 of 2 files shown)


src/assignment/views/AssignmentEdit.tsx
```diff
@@ -891,7 +891,9 @@ const AssignmentEdit: FunctionComponent<AssignmentEditProps & UserProps> = ({
 
 					<InActivityWarningModal
 						onActivity={updateAssignmentEditor}
-						onExit={releaseAssignmentEditStatus}
+						// AVO-2934: do not release edit status after delete assignment
+						//onExit={releaseAssignmentEditStatus}
+						onExit={() => true}
```
bertyhell, line 896: shouldn't this still do this if the assignment wasn't deleted?

eg:
if assignment was deleted:
    exit () => true
else 
    releaseAssignmentEditStatus

PR #1915, opened 2023-12-11
fix(avo-2738): fix query

WardVerc1:
image

Ticket: https://meemoo.atlassian.net/browse/AVO-2738


src/admin/assignments/views/AssignmentsOverviewAdmin.tsx
```diff
@@ -198,6 +194,8 @@ const AssignmentOverviewAdmin: FunctionComponent<RouteComponentProps & UserProps
 		}
 	}, [fetchAssignments, assignments]);
 
+	//console.log(tableState);
```
bertyhell, line 197: remove

bertyhell (approved):

PR #1927, opened 2023-12-18
fix(avo-2918): fix header for interactive tour

WardVerc1:
image

image

image

image

Ticket: https://meemoo.atlassian.net/browse/AVO-2918

### DIFF
--- a/src/assignment/views/AssignmentDetail.scss
+++ b/src/assignment/views/AssignmentDetail.scss
@@ -8,3 +8,7 @@
 		margin-left: 10.6rem; // = distance assignment-item icon + paddings
 	}
 }
+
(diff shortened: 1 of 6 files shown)


bertyhell:
maybe we should handle this in a more general way: https://github.com/viaacode/avo2-client/pull/1933

so we don't have to add padding to all header components bottom rows
and we also avoid using the spacer components since those are bad practice according to modern web development

This pr moves all interactive tours in the bottom right row of the header down. So the header component itself still works as expected with padding.

bertyhell (commented):
see global remark.

WardVerc1:
Other PR has better solution

PR #1932, opened 2023-12-22
fix(avo-3047): make theme not required for secondary

WardVerc1:
Ticket: https://meemoo.atlassian.net/browse/AVO-3047


WardVerc1:
Wrong solution, must be checked which teacher value was filled in in the form

PR #1972, opened 2024-03-13
fix(AVO-3132): split copy link from title link for coll, bun. & assign.

bertyhell:
https://meemoo.atlassian.net/browse/AVO-3132

* make one render function for rendering titles and kopie label
* unnest the link for the title from the link of the copy label
* use function on pages
    * assignment overview
    * collection/bundle overview
    * collection/bundle actualisation overview
    * collection/bundle quality overview
    * collection/bundle communication overview
* replace usage of lodash get function with elvis operator ?.
* replace ReactText with ReactNode
* fix error in author user group filter property: tableState.author.user_groups => tableState.author_user_group

image

### DIFF
--- a/src/shared/helpers/render-collection-or-bundle-or-assignment-title-and-copy-tag.tsx
+++ b/src/shared/helpers/render-collection-or-bundle-or-assignment-title-and-copy-tag.tsx
@@ -1,0 +1,39 @@
+import { TagList } from '@viaa/avo2-components';
+import type { Avo } from '@viaa/avo2-types';
+import { truncate } from 'lodash-es';
+import React from 'react';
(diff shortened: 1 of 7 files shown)


src/shared/helpers/render-collection-or-bundle-or-assignment-title-and-copy-tag.tsx
ian-emsens-shd, line 1: This is a component, not a helper. Could also choose to pull the entire switch/case out of this view into a separate component. See `src/shared/components/QuickLaneFilterTableCell/QuickLaneFilterTableCell.tsx`

ian-emsens-shd (changes requested):

PR #1979, opened 2024-04-24
Feature/avo 3165 skip terms of service for elementary pupils

ian-emsens-shd:
- feat(AVO-3165): Don't show terms of service to elementary pupils + scaffold landingpage


src/shared/helpers/user-is.ts
```diff
@@ -0,0 +1,15 @@
+import { CommonUserSchema } from '@viaa/avo2-types/types/user';
+
+import { SpecialUserGroup } from '../../admin/user-groups/user-group.const';
+
+export function userIsPupil(user: CommonUserSchema) {
+	return userIs(user, SpecialUserGroup.Pupil);
```
bertyhell, line 6: Will this one still be used? Shouldn't all pupils either be ElementaryPupil or SecondaryPupil after the migration?

bertyhell (approved):

src/authentication/views/registration-flow/accept-elementary-pupil-conditions.tsx
```diff
@@ -0,0 +1,102 @@
...
+			</Container>
+			<Spacer margin="large">
+				<Toolbar>
+					<ToolbarCenter>
+						{loading ? (
+							<Spinner size={'large'} />
+						) : (
+							<Button
+								label={tText('TODO')}
+								title={tText('TODO')}
```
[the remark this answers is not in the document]
ian-emsens-shd: Out of scope for this ticket, https://github.com/viaacode/avo2-client/pull/1981

src/shared/helpers/user-is.ts
ian-emsens-shd: User group was renamed, not created again. `SpecialUserGroup.Pupil` will refer to secondary pupils post-migration.

bertyhell: ok, so maybe we should also rename the name of this entry in the enum, but keep the value the same?
```SpecialUserGroup.PupilSecundary```?

bertyhell: Ok, then we should probably rename the enum to PupilSecundary but leave its underlaying value the same

PR #1985, opened 2024-05-06
Feature/avo 3179 determine education level on copy

ian-emsens-shd:
| Kenmerk opdracht | Onderwijsniveau lesgever | Resultaat op kopie |
|:----------------:|:------------------------:|:------------------:|
| Secundair        | Secundair                | Secundair          |
| Secundair        | Lager                    | Lager              |
| Lager            | Lager                    | Lager              |
| Lager            | Secundair                | Secundair          |
| Secundair        | Lager en Secundair       | Secundair          |
| Lager            | Lager en Secundair       | Lager              |


src/assignment/assignment.service.ts
```diff
@@ -1,5 +1,6 @@
 import { fetchWithLogoutJson } from '@meemoo/admin-core-ui';
 import { type Avo } from '@viaa/avo2-types';
+import { type UserSchema } from '@viaa/avo2-types/types/user';
```
bertyhell, line 3: Don't use deep imports
import Avo
```suggestion
import { type Avo } from '@viaa/avo2-types';
```

And then use this instead of UserSchema. You should never use types with Schema at the end. That's the internal naming for the typings repo
```
Avo.User.User
```

src/assignment/assignment.service.ts
```diff
@@ -509,21 +516,36 @@ export class AssignmentService {
 	static async duplicateAssignment(
 		newTitle: string,
 		initialAssignment: Partial<Avo.Assignment.Assignment> | null,
-		profileId: string
+		user: UserSchema
 	): Promise<Avo.Assignment.Assignment> {
-		if (!initialAssignment || !initialAssignment.id) {
+		const owner_profile_id = user.profile?.id;
```
bertyhell, line 521: The code should use camelcase for variables
only switch to snake case at the last moment just before interacting with the database

```suggestion
		const ownerProfileId = user.profile?.id;
```

src/assignment/helpers/duplicate-assignment.ts
```diff
@@ -7,7 +8,7 @@ import { AssignmentService } from '../assignment.service';
 
 export async function duplicateAssignment(
 	assignment: Avo.Assignment.Assignment | null | undefined,
-	profileId: string
+	user: UserSchema
```
bertyhell, line 11: no deep imports

src/shared/helpers/is-user.ts
```diff
@@ -4,19 +4,21 @@ import { SpecialUserGroup } from '../../admin/user-groups/user-group.const';
 
 import { EducationLevelId } from './lom';
 
-export function isUserPupil(user: CommonUserSchema) {
+type User = Pick<CommonUserSchema, 'loms' | 'userGroup'>;
```
bertyhell, line 7:
```suggestion
type UserLomsAndUserGroup = Pick<CommonUserSchema, 'loms' | 'userGroup'>;
```

bertyhell (approved):

PR #1996, opened 2024-05-15
Feature/avo 3183 tracking events

ian-emsens-shd:


src/assignment/assignment.service.ts
```diff
@@ -1480,10 +1481,15 @@ export class AssignmentService {
...
+		assignment: Avo.Assignment.Assignment | null | undefined,
+		invitee: Partial<ContributorInfo>,
+		inviter?: Avo.User.CommonUser
 	): Promise<void> {
-		if (isNil(user.email) || isEmpty(user.email)) {
+		if (!assignment) return;
+
+		const assignmentId = assignment.id;
+
+		if (isNil(invitee.email) || isEmpty(invitee.email)) {
```
bertyhell, line 1492: !variable does the same as combining isNil and isEmpty and is a lot more efficient
```suggestion
		if (!invitee.email) {
```

src/assignment/assignment.service.ts
```diff
@@ -1492,16 +1498,30 @@ export class AssignmentService {
...
+					object: assignmentId,
+					object_type: 'assignment',
+					action: 'share',
+					resource: {
+						education_level: assignment.education_level_id,
+						rights: invitee.rights || 'NONE',
+					},
+				},
+				inviter
+			);
```
bertyhell, line 1519: check that these events make it into the events database
https://event-logger-qas-meemoo-infra.private.cloud.meemoo.be/console/api/api-explorer

password in 1p

src/assignment/views/AssignmentDetail.tsx
```diff
@@ -506,7 +509,8 @@ const AssignmentDetail: FC<
 	};
 
 	const onDeleteAssignment = async (): Promise<void> => {
-		await deleteAssignment(assignmentId, user, () =>
+		if (!assignment) return;
```
bertyhell, line 512: add a toast message: Failed to delete the assignment, reload the page and try again

src/assignment/helpers/delete-assignment.ts
```diff
@@ -39,6 +41,9 @@ export async function deleteAssignment(
 				object: assignmentId,
 				object_type: 'assignment',
 				action: 'delete',
+				resource: {
+					education_level: String(assignment?.education_level_id),
```
bertyhell, line 45: why are these casts to String needed?
This will also cause the value to be the string 'undefined' if it is undefined
Why not just leave it as what it was without the cast?

same remark for all events

bertyhell (approved):

src/assignment/assignment.service.ts
ian-emsens-shd: Yup, being saved correctly see `a4d791c6-eacc-44cc-9628-74144ca3be24`

src/assignment/helpers/delete-assignment.ts
ian-emsens-shd: `resource` is typed as `Record<string, string | string[] | boolean | number | Avo.Search.DateRange>`, `undefined` is none of ` string | string[] | boolean | number | Avo.Search.DateRange`.

PR #2002, opened 2024-05-28
Feature/avo 3195 e2e share assignment

ian-emsens-shd:
```
% npm run test:e2e:debug:t65

> tests@1.0.0 test:e2e:debug:t65
> cross-env PWDEBUG=console dotenv -e ../.env.qas npx playwright test tests/t65-opdracht-delen.spec


Running 4 tests using 1 worker
  4 passed (2.1m)
```

### DIFF
--- a/tests/tests/t65-opdracht-delen.spec.ts
+++ b/tests/tests/t65-opdracht-delen.spec.ts
@@ -1,0 +1,211 @@
+import { expect, type Page, test } from '@playwright/test';
+
+import { cleanupTestdata } from '../helpers/cleanup';
+import { goToPageAndAcceptCookies } from '../helpers/go-to-page-and-accept-cookies';
+import { loginOnderwijsAvo } from '../helpers/login-onderwijs-avo';
+import { logoutOnderwijsAvo } from '../helpers/logout-onderwijs-avo';
+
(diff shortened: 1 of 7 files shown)


tests/tests/t65-opdracht-delen.spec.ts
```diff
@@ -0,0 +1,211 @@
...
+		},
+		{
+			user: process.env.TEST_LESGEVER_BEIDE_GEBRUIKER_USER as string,
+			pass: process.env.TEST_LESGEVER_BEIDE_GEBRUIKER_PASS as string,
+		},
+		true
+	);
+});
+
+async function createAssignmentWithTitle(page: Page) {
```
bertyhell, line 81: shouldn't these helpers be moved to the helpers folder
or will they only be used inside this test?

bertyhell (approved):

tests/tests/t65-opdracht-delen.spec.ts
ian-emsens-shd: Lots of tests use slightly similar code. I split it into declarative functions in this file for my own sanity but I don't think we can easily implement these as helpers in other files.

PR #2004, opened 2024-05-29
Feature/avo 3188 sticky assignment heading

ian-emsens-shd:
fix(AVO-3188): make pupil heading sticky and condensed

### DIFF
--- a/src/assignment/components/AssignmentHeading.tsx
+++ b/src/assignment/components/AssignmentHeading.tsx
@@ -1,3 +1,6 @@
+import './AssignmentHeading.scss';
+
+import { useWindowScroll } from '@uidotdev/usehooks';
 import {
 	Container,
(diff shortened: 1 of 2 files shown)


src/assignment/components/AssignmentHeading.tsx
```diff
@@ -6,13 +9,11 @@ import {
 	ToolbarLeft,
 	ToolbarRight,
 } from '@viaa/avo2-components';
-import classnames from 'classnames';
+import classNames from 'classnames';
```
bertyhell, line 12: prefer the migration towards clsx
its a smaller package and it is a drop in replacement for classnames
it should be already available in the avo-client repo
```import clsx from 'clsx';```

bertyhell (approved):

PR #2008, opened 2024-05-29
bugfix/AVO-3269-filter-subjects-by-level-for-assignments

ian-emsens-shd:
fix(AVO-3269): add filtering for assignment meta subjects

### DIFF
--- a/src/assignment/components/AssignmentMetaDataFormEditable.tsx
+++ b/src/assignment/components/AssignmentMetaDataFormEditable.tsx
@@ -17,6 +17,7 @@ import { type UseFormSetValue } from 'react-hook-form';
 
 import { ShortDescriptionField, ThumbnailStillsModal } from '../../shared/components';
 import LomFieldsInput from '../../shared/components/LomFieldsInput/LomFieldsInput';
+import { EducationLevelType } from '../../shared/helpers/lom';
 import { tHtml } from '../../shared/helpers/translate';
 import useTranslation from '../../shared/hooks/useTranslation';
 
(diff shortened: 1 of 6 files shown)


src/assignment/components/AssignmentMetaDataFormEditable.tsx
```diff
@@ -56,6 +56,19 @@ const AssignmentMetaDataFormEditable: FC<AssignmentMetaDataFormEditableProps> =
 		}));
 	};
 
+	const filterSubjects = (subject: Avo.Lom.LomField & { related?: string[] }) => {
+		const selectedEducationLevels = (assignment.loms || []).filter(({ lom }) => {
+			return lom?.scheme === 'https://w3id.org/onderwijs-vlaanderen/id/structuur';
```
bertyhell, line 61: use an enum value for this since it is used multiple times in the codebase already

src/shared/components/LomFieldsInput/LomFieldsInput.tsx
```diff
@@ -56,6 +57,7 @@ const LomFieldsInput: FC<LomFieldsInputProps> = ({
 	subjectsPlaceholder,
 	themesPlaceholder,
 	limitDegreesByAlreadySelectedLevels = false,
+	filterSubjects = () => true,
```
bertyhell, line 60: if the response is always true, then there is no point of having the filter function
```suggestion
	filterSubjects = () => boolean,
```

bertyhell (approved):

src/shared/components/LomFieldsInput/LomFieldsInput.tsx
ian-emsens-shd: This is the default value, typing is set to `filterSubjects?: (subject: Avo.Lom.LomField) => boolean;` above.

PR #2014, opened 2024-06-04
Bugfix/avo 3278 add missing where

ian-emsens-shd:


src/admin/assignments/assignments.types.ts
```diff
@@ -25,6 +25,7 @@ export interface AssignmentsOverviewTableState extends FilterableTableState {
 	share_type: string;
 	subjects: string[];
 	education_levels: string[];
+	education_level: string[]; // "Kenmerk"
```
bertyhell, line 28: maybe add a bit more comments here
```suggestion
	education_levels: string[]; // These are the lom values for the assignment publication details
	education_level: string[]; // This is the type of assignment: Lager onderwijs or secundair onderwijs. also known as "Kenmerk"
```

src/admin/assignments/views/AssignmentsOverviewAdmin.tsx
```diff
@@ -296,6 +297,12 @@ const AssignmentOverviewAdmin: FunctionComponent<RouteComponentProps & UserProps
 			);
 		}
 
+		if (filters.education_level && filters.education_level) {
```
bertyhell, line 300: the same condition twice?

src/admin/shared/helpers/filters.ts
```diff
@@ -264,3 +264,35 @@ export function generateLomFilter(values: string[], scheme: EducationLevelType):
 		};
 	}
 }
+
+/**
+ * Does the same thing as `generateLomFilter` but matches a different structure
+ */
+export function generateEducationLevelFilter(values: string[], scheme: EducationLevelType): any {
```
bertyhell, line 271:
```suggestion
export function generateEducationLevelFilter(educationLevels: string[], scheme: EducationLevelType): any {
```

bertyhell (approved):

PR #2016, opened 2024-06-05
Bugfix/avo 3284 skip warning for viewers

ian-emsens-shd:
See recording attached to https://meemoo.atlassian.net/browse/AVO-3284


src/shared/components/ShareWithColleagues/ShareWithColleagues.tsx
```diff
@@ -129,7 +129,7 @@ const ShareWithColleagues: FC<ShareWithColleaguesProps & UserProps> = ({
 	};
 
 	const handleAddNewContributor = () => {
-		if (isShareWarningModalRemembered) {
+		if (isShareWarningModalRemembered || contributor.rights === ContributorInfoRight.VIEWER) {
```
bertyhell, line 132: maybe also add some comments here, we you check these 2 values

bertyhell (approved):

PR #2038, opened 2024-07-01
feat(ARC-2096): integrate multilanguage admin-core and hide language

bertyhell:
https://meemoo.atlassian.net/browse/ARC-2096

requires admin-core PR: https://github.com/viaacode/react-admin-core-module/pull/265

* Integrate the multilanguage admin-core into the client
* hide all language fields using a config option: config.components.enableMultiLanguage

image
image
image
image
image
image
image
image
image
image
image
image


src/admin/admin.const.ts
```diff
@@ -114,7 +115,7 @@ async function getContentPageDetailRouteByPath(
 ): Promise<string | undefined> {
 	try {
 		const page = await ContentPageService.getContentPageByLanguageAndPath(
-			LanguageCode.Nl,
+			Locale.Nl as any,
```
ian-emsens-shd, line 118: Avoid cast

ian-emsens-shd (approved):

PR #2044, opened 2024-07-11
feat(AVO-3327): add hook and implement for elementary pupil conditions

ian-emsens-shd:
Note: this nearly drove me insane today

### DIFF
--- a/src/shared/hooks/useDisablePictureInPicture.tsx
+++ b/src/shared/hooks/useDisablePictureInPicture.tsx
@@ -1,0 +1,20 @@
+import { type RefObject, useEffect } from 'react';
+
+export const useDisablePictureInPicture = (ref: RefObject<HTMLElement>) => {
+	useEffect(() => {
+		if (ref.current) {
+			observer.observe(ref.current, {
+				childList: true,
+				subtree: true,
+			});
+		}
+
+		return () => observer.disconnect();
+	});
+
+	const observer = new MutationObserver(() => {
+		Array.from(document.getElementsByTagName('video')).forEach((element) => {
+			element.disablePictureInPicture = true;
+		});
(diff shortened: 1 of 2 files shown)


src/shared/hooks/useDisablePictureInPicture.tsx
```diff
@@ -0,0 +1,20 @@
...
+		}
+
+		return () => observer.disconnect();
+	});
+
+	const observer = new MutationObserver(() => {
+		Array.from(document.getElementsByTagName('video')).forEach((element) => {
+			element.disablePictureInPicture = true;
+		});
+	});
```
bertyhell, line 19: Seems more logical to define the observer const first, then use it in the useEffect
```suggestion
	const observer = new MutationObserver(() => {
		Array.from(document.getElementsByTagName('video')).forEach((element) => {
			element.disablePictureInPicture = true;
		});
	});

	useEffect(() => {
		if (ref.current) {
			observer.observe(ref.current, {
				childList: true,
				subtree: true,
			});
		}

		return () => observer.disconnect();
	});
```

src/shared/hooks/useDisablePictureInPicture.tsx
```diff
@@ -0,0 +1,20 @@
...
+export const useDisablePictureInPicture = (ref: RefObject<HTMLElement>) => {
+	useEffect(() => {
+		if (ref.current) {
+			observer.observe(ref.current, {
+				childList: true,
+				subtree: true,
+			});
+		}
+
+		return () => observer.disconnect();
```
bertyhell, line 12: Does this also work?
```suggestion
		return observer.disconnect;
```

bertyhell (approved):

PR #2047, opened 2024-09-10
Feature/avo 3340 cut quick lane items

ian-emsens-shd:


src/quick-lane/views/QuickLaneDetail.tsx
```diff
@@ -205,11 +212,14 @@ const QuickLaneDetail: FunctionComponent<QuickLaneDetailProps> = ({
...
-						verticalLayout={isMobileWidth()}
-					/>
+					<>
+						<ItemVideoDescription
+							itemMetaData={quickLane.content as Avo.Item.Item}
+							showDescription={contentLayout === AssignmentLayout.PlayerAndText}
+							verticalLayout={isMobileWidth()}
+							cuePointsLabel={{ start, end }}
+						/>
+					</>
```
bertyhell, line 222: do we need the fragment wrapper <></> here?

src/shared/components/QuickLaneModal/QuickLaneModalSharingTab.tsx
```diff
@@ -34,9 +39,20 @@ const QuickLaneModalSharingTab: FunctionComponent<QuickLaneModalProps & UserProp
...
+	);
+
 	const [quickLane, setQuickLane] = useState<QuickLaneUrlObject>(defaultQuickLaneState);
 	const [exists, setExists] = useState<boolean>(false);
 	const [synced, setSynced] = useState<boolean>(false);
+	const [fragmentStartTime, setFragmentStartTime] = useState<number>(
+		quickLane.start_oc || defaultQuickLaneState.start_oc || 0
+	);
+	const [fragmentEndTime, setFragmentEndTime] = useState<number>(
+		quickLane.end_oc || fragmentDuration || defaultQuickLaneState.end_oc || 0
```
bertyhell, line 54: be careful that the end time can never be before the start time
or the range component in the time crop controls will throw an exception that causes the page to crash
But maybe this is already checked inside the time crop controls component?

bertyhell (approved):

src/quick-lane/views/QuickLaneDetail.tsx
ian-emsens-shd: Nope, will remove

src/shared/components/QuickLaneModal/QuickLaneModalSharingTab.tsx
ian-emsens-shd: Will check

PR #2065, opened 2024-10-09
feat(AVO-3249): switch assignments overview to proxy endpoint

bertyhell:
https://meemoo.atlassian.net/browse/AVO-3249

Requires proxy PR: https://github.com/viaacode/avo2-proxy/pull/771

instead of a client graphql query
This reduces codes in the client
Also used react-query to fetch the assignments, so better handling of loading states

https://github.com/user-attachments/assets/d5da4c59-0cbf-4b56-a702-2243d0000b27


src/assignment/hooks/useGetAssignments.tsx
```diff
@@ -0,0 +1,41 @@
...
+		sortColumn: AssignmentOverviewTableColumns;
+		sortOrder: Avo.Search.OrderDirection;
+		tableColumnDataType: TableColumnDataType;
+		offset: number;
+		limit?: number;
+		filterString: string | undefined;
+		labelIds: string[] | undefined;
+		classIds: string[] | undefined;
+		shareTypeIds: string[] | undefined;
+	},
```
ian-emsens-shd, line 21: Able to (partially) share types with `AssignmentService.fetchAssignments`?

src/assignment/modals/ImportToAssignmentModal.tsx
```diff
@@ -103,20 +104,19 @@ const ImportToAssignmentModal: FunctionComponent<ImportToAssignmentModalProps &
...
-				[],
-				null // limit: no limit
-			);
+				tableColumnDataType: columnDataType,
+				offset: 0,
+				limit: ITEMS_PER_PAGE,
+				filterString,
+				labelIds: [],
+				classIds: [],
+				shareTypeIds: [],
```
ian-emsens-shd, line 118: Still need to pass empty arrays here? Looks like they're set to default to `undefined` in the data call

src/assignment/views/AssignmentOverview.tsx
```diff
@@ -134,8 +127,24 @@ const AssignmentOverview: FunctionComponent<
 	const [markedAssignment, setMarkedAssignment] = useState<Avo.Assignment.Assignment | null>(
 		null
 	);
-	const [canEditAssignments, setCanEditAssignments] = useState<boolean | null>(null);
-	const [showPublicState, setShowPublicState] = useState<boolean | null>(null);
+	const canEditAssignments: boolean | null = useMemo(() => {
```
ian-emsens-shd, line 130: Prefer type inference like `showPublicState`

src/assignment/views/AssignmentOverview.tsx
```diff
@@ -134,8 +127,24 @@ const AssignmentOverview: FunctionComponent<
...
+		if (!commonUser) {
+			return null;
+		}
+		return (
+			PermissionService.hasPerm(commonUser, PermissionName.EDIT_ANY_ASSIGNMENTS) ||
+			PermissionService.hasPerm(commonUser, PermissionName.EDIT_OWN_ASSIGNMENTS)
+		);
```
ian-emsens-shd, line 137: Not sure we ever want to handle null (loading?) states for permissions, if the user data isn't loaded that means they can't do permission-enabled actions imo.

```suggestion
		return commonUser && (
			PermissionService.hasPerm(commonUser, PermissionName.EDIT_ANY_ASSIGNMENTS) ||
			PermissionService.hasPerm(commonUser, PermissionName.EDIT_OWN_ASSIGNMENTS)
		);
```

src/assignment/views/AssignmentOverview.tsx
```diff
@@ -1006,8 +947,12 @@ const AssignmentOverview: FunctionComponent<
...
 
 	const renderAssignmentsView = () => {
-		if (!assignments) {
-			return null;
+		if (isLoadingAssignments) {
+			return (
+				<Flex orientation="horizontal" center>
+					<Spinner size="large" />
+				</Flex>
+			);
```
ian-emsens-shd, line 955: Not using `LoadingErrorLoadedComponent`?

ian-emsens-shd (changes requested):

src/assignment/views/AssignmentOverview.tsx
bertyhell: No, i'm fasing that component out, since it's better handled by react-query functions loading states
and already showing some ui before all the data loads

src/assignment/hooks/useGetAssignments.tsx
bertyhell: yes, good idea 👍

bertyhell:
fixed remarks in fd188bd52c0e5f41603839137cdae05af89ba071

ian-emsens-shd (approved):

PR #2068, opened 2024-10-11
fix(admin): add commonUser prop to admin collection screens

bertyhell:


ian-emsens-shd (approved):

bertyhell:
fixed and merged into develop

PR #2070, opened 2024-10-11
fix(AVO-2982): change close icon to cross instead of eye

bertyhell:
https://meemoo.atlassian.net/browse/AVO-2982

this used to be "view" to go back to the detail page

image
image


src/admin/assignments/assignments.const.ts
```diff
@@ -187,7 +187,7 @@ export const GET_ASSIGNMENT_OVERVIEW_TABLE_COLS = (
 		},
 		{
 			id: 'education_degrees',
-			label: tText('Onderwijsgraden'),
+			label: tText('admin/assignments/assignments___onderwijsgraden'),
```
ian-emsens-shd, line 190: ✅

ian-emsens-shd (approved):

PR #2071, opened 2024-10-14
feat(AVO-3401): open duplicated assignment in new tab

bertyhell:
https://meemoo.atlassian.net/browse/AVO-3401

image
image


ian-emsens-shd (approved):

bertyhell:
fixed and merged into develop

PR #2073, opened 2024-10-25
feature: AVO-1827 event tracking

bertyhell:
https://meemoo.atlassian.net/browse/AVO-1827

requires database branch: https://github.com/viaacode/hasura-graphql/tree/feature/AVO-1827-view-and-play-events
requires proxy PR: https://github.com/viaacode/avo2-proxy/pull/774


src/shared/services/event-logging-service.ts
ian-emsens-shd, line 1: Not code related but should this service comply with cookie preferences?

ian-emsens-shd (approved):

src/shared/services/event-logging-service.ts
bertyhell: i don't think so. Isn't the cookie declaration mostly about what we store on the user's device. Like cookies and local storage.

But this does have to follow the GDPR regulations about what we track about a user.
But i think meemoo adds a clause about being able to store info and contact the user in relation to a properly working platform. So i think we're covered

PR #2076, opened 2024-10-30
CI/CD

ian-emsens-shd:
https://meemoo.atlassian.net/browse/AVO-2462
Work in progress, fixes some and skips most E2E tests


bertyhell:
you know you can run test separatly, right? So the skipping isn't really nessesary until all separate tests have been fixed.
We should also strive for all tests to be independant, so you never have to run tast A so that B can succeed.
That way we can just fix al separate tests, and then start testing them all at once.
https://github.com/viaacode/avo2-client/blob/e36a121db5dc5d4b915f5c066114d857b9c5e593/tests/package.json#L15

PR #2086, opened 2025-03-18
AVO-3616: load klascement publish info and set fields readonly when a…

reunefe:
- Load publish info on start
- refresh publish info after publish
- open new window to klascement if publish was successful
- set klascement form to readonly when published succesfully


.env.example
```diff
@@ -38,3 +38,5 @@ TEST_LESGEVER_LAGER_GEBRUIKER_USER=
 TEST_LESGEVER_LAGER_GEBRUIKER_PASS=
 TEST_LESGEVER_SECUNDAIR_GEBRUIKER_USER=
 TEST_LESGEVER_SECUNDAIR_GEBRUIKER_PASS=
+
+KLASCEMENT_URL=https://www.klascement.net
```
bertyhell, line 42: put this env var above the TEST env vars, since the TEST env vars are not getting injected into the client
and we want to keep all injected env vars together

Also make sure you update the 1password env vars voor de client when you change them
`hetarchief - client - .env (local)`

src/collection/components/CollectionOrBundleEditMarcom.tsx
```diff
@@ -81,6 +83,11 @@ const CollectionOrBundleEditMarcom: FC<CollectionOrBundleEditMarcomProps & UserP
 	const { mutateAsync: publishCollectionToKlascement, isLoading: isPublishing } =
 		usePublishCollectionToKlascement();
 
+	const isPublishedToKlascement = useMemo(
+		() => klascementId !== undefined && klascementId !== null,
```
bertyhell, line 87: optioneel: je kan ook de isNil() helper van lodash gebruiken

src/collection/components/CollectionOrBundleEditMarcom.tsx
```diff
@@ -98,6 +105,25 @@ const CollectionOrBundleEditMarcom: FC<CollectionOrBundleEditMarcomProps & UserP
...
+			setKlascementId(publishInfo?.klascement_id ?? undefined);
+		} catch (err) {
+			console.error(
+				new CustomError('Failed to fetch klascement publish info from the database', err, {
+					collectionId: collection.id,
+				})
+			);
+			ToastService.danger(tHtml('Het ophalen van de klascement informatie is mislukt'));
+		}
+	}, [collection.id, tHtml]);
```
bertyhell, line 125: Wrap this in a react-query wrapper so we get loading states, error states and caching
```
const {data: publishInfo, refetch: refetchPublishInfo, isLoading: isLoadingPublishInfo} = useGetKlascementPublishInfo(collection.id);
```

```
export const useGetKlascementPublishInfo = (collectionId: string): UseQueryResult<KlascementPublishInfo> => {
	return useQuery(
		[QUERY_KEYS.GET_KLASCEMENT_PUBLISH_INFO, collectionId],
		() => {
			return KlascementService.getKlascementPublishInfoForCollection(collectionId);
		}
	);
};
```

bertyhell (approved):

.env.example
reunefe: Moved upwards and added to 1Password

src/collection/components/CollectionOrBundleEditMarcom.tsx
reunefe: Always forget about that one

PR #2090, opened 2025-04-01
#AVO-3636: Combine current share options for fragments to single modal

reunefe:
- Move content of `ShareThroughEmailModal` to its own component (`ShareThroughEmailContent`)
- Use `ShareThroughEmailContent` in `ShareThroughEmailModal`
- Move content of `QuickLaneModal` to its own component (`QuickLaneContent`)
- Use `QuickLaneContent` in `QuickLaneModal`
- Create new `FragmentShareModal` containing tabs to share with colleagues or pupils and use the newly created components
- Remove double share buttons on a fragment and add 1 new button to open the new `FragmentShareModal`


src/shared/components/QuickLaneContent/QuickLaneContent.tsx
```diff
@@ -0,0 +1,235 @@
...
+import QuickLaneContentPublicationTab from './QuickLaneContentPublicationTab';
+import QuickLaneContentSharingTab from './QuickLaneContentSharingTab';
+import './QuickLaneContent.scss';
+
+// State
+
+const QuickLaneContentTabs = {
+	publication: 'publication',
+	sharing: 'sharing',
+};
```
bertyhell, line 23: Isn't an enum more suitable?
```suggestion
enum QuickLaneContentTabs {
	publication = 'publication',
	sharing = 'sharing',
}
```

src/shared/components/QuickLaneContent/QuickLaneContent.tsx
```diff
@@ -0,0 +1,235 @@
...
+	const getTabs = () => {
+		// AVO-1880
+		if ((content as Avo.Collection.Collection).is_public) {
+			return [];
+		}
+
+		return tabs.filter((tab) => {
+			switch (tab.id) {
+				case QuickLaneContentTabs.publication:
+					return content_label === 'COLLECTIE' && canPublish;
```
bertyhell, line 125: maybe store the `content_label === 'COLLECTIE'` in a computed variable, since it's reused a lot
```typescript
const isCollection = content_label === 'COLLECTIE';
```

bertyhell (approved):

src/shared/components/QuickLaneContent/QuickLaneContent.tsx
reunefe: I just split the existing code, without changing anything. But indeed, I added an enum for these checks and stored the check in a variable

PR #2092, opened 2025-04-22
#AVO-3650: start embeds for workspace

reunefe:


src/shared/services/embed-code-service.ts
```diff
@@ -56,4 +39,31 @@ export class EmbedCodeService {
...
+	}> {
+		let url: string | undefined = undefined;
+
+		try {
+			url = stringifyUrl({
+				url: `${getEnv('PROXY_URL')}/embed-codes`,
+				query: {
+					sortColumn: params?.sortColumn,
+					sortOrder: params?.sortOrder,
+					tableColumnDataType: params?.sortType,
```
bertyhell, line 55: if we build the order object on the proxy then we don't need to pass this type to the proxy
```suggestion
```

src/workspace/views/EmbedCodeOverview.tsx
```diff
@@ -0,0 +1,312 @@
...
+// Typings
+interface EmbedCodeOverviewProps {
+	numberOfItems: number;
+	onUpdate: () => void | Promise<void>;
+}
+
+enum EmbedCodeAction {
+	EDIT = 'EDIT',
+	COPY = 'COPY',
+	DUPLICATE = 'DUPLICATE',
```
bertyhell, line 36: what's the difference between copy and duplicate?
maybe rename COPY => COPY_TO_CLIPBOARD
```suggestion
	COPY_TO_CLIPBOARD= 'COPY_TO_CLIPBOARD',
	DUPLICATE = 'DUPLICATE',
```

bertyhell (approved):

bertyhell:
Did you test the mobile view of this table?
You can use the mobile view of the collections or assignments as inspiration

PR #2094, opened 2025-05-05
#AVO-3658: Add embed modal to bookmarks + move load bookmarks to API …

reunefe:
…instead of having the query in the FE


src/workspace/views/BookmarksOverview.tsx
```diff
@@ -244,18 +258,55 @@ const BookmarksOverview: FC<BookmarksOverviewProps & UserProps & RouteComponentP
...
+									duration: bookmarkInfo.contentDuration,
+									external_id: bookmarkInfo.contentLinkId,
+									thumbnail_path: bookmarkInfo.contentThumbnailPath,
+									type: bookmarkInfo.contentType,
+								},
+								descriptionType: EmbedCodeDescriptionType.ORIGINAL,
+								description: bookmarkInfo.contentDescription,
+								start: 0,
+								end: toSeconds(bookmarkInfo.contentDuration),
+							} as EmbedCode);
```
bertyhell, line 303: maybe move this to handleEmbedCodeClicked handler to make the html more clear

src/workspace/views/BookmarksOverview.tsx
```diff
@@ -335,6 +386,11 @@ const BookmarksOverview: FC<BookmarksOverviewProps & UserProps & RouteComponentP
 	const renderBookmarks = () => (
 		<>
 			{bookmarks && bookmarks.length ? renderTable() : renderEmptyFallback()}
+			<BookmarkEmbedModal
+				isOpen={!!isEmbedCodeModalOpen}
+				embedCode={isEmbedCodeModalOpen}
+				onClose={() => setIsEmbedCodeModalOpen(undefined)}
```
bertyhell, line 392: PERSONAL PREFERENCE:
for intentional clearing of variables i prefer null
things that are accidentally undefined i prefer undefined
```suggestion
				onClose={() => setIsEmbedCodeModalOpen(null)}
```

src/workspace/workspace.service.ts
```diff
@@ -16,4 +17,16 @@ export class WorkspaceService {
 			throw new CustomError('Failed to get workspace counts', err, { url });
 		}
 	}
+
+	public static getAllBookmarksForUser(): Promise<BookmarkInfo[]> {
+		let url: string | null = null;
+		try {
+			url = (getEnv('PROXY_URL') as string) + '/workspace/bookmarksForUser';
```
bertyhell, line 24: Most paths in the avo proxy use kebabcase for their routes
Also the "ForUser" seems quite obvious to me, since we're inside the workspace route
So i would switch it to:
```suggestion
			url = (getEnv('PROXY_URL') as string) + '/workspace/bookmarks';
```

bertyhell (approved):

PR #2105, opened 2025-05-27
#AVO-3684: 

reunefe:
store embed-flow when opening from another window in redux store + show smartschool button instead of share button when embed-flow is set to smartschool.

Quick check https://codepen.io/ikkimausi/pen/GgJgvRy


src/App.tsx
```diff
@@ -142,6 +143,18 @@ const App: FC<
 		}
 	}, [location, props]);
 
+	/**
+	 * Check if this window was opened from somewhere else and get the embed-flow
+	 * Store them in the redux store
+	 */
```
bertyhell, line 149:
```suggestion
	/**
	 * Check if this window was opened from somewhere else and get the embed-flow query param
	 * Store the embed-flow query param in the redux store
	 */
```

bertyhell (approved):

bertyhell:
Should we still show the bookwidgets tab in the share modal when embed-flow===smartschool?
Or should we hide it?
image

I'm guessing this isn't specified in the FA?

reunefe:
> Should we still show the bookwidgets tab in the share modal when embed-flow===smartschool? Or should we hide it? image
> 
> I'm guessing this isn't specified in the FA?

I'll double check, I did not check this. Just built the create flow but forget the workspace flows 🙈

PR #2107, opened 2025-06-03
#AVO-3679: introduce new property to skip the poster to init the flow…

reunefe:
…player without auto starting the video

### DIFF
--- a/src/embed/components/Embed.tsx
+++ b/src/embed/components/Embed.tsx
@@ -81,7 +81,7 @@ const Embed: FC<UserProps> = ({ commonUser }) => {
 				poster={getFlowPlayerPoster(undefined, embedCode.content as ItemSchema)}
 				item={embedCode.content as ItemSchema}
 				canPlay={true}
-				initFlowPlayerWithoutAutoPlay={true}
+				placeholder={false}
 				cuePointsLabel={{ start: embedCode.start, end: embedCode.end }}
 				cuePointsVideo={{ start: embedCode.start, end: embedCode.end }}
 				onPlay={onPlay}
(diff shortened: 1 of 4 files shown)


src/embed/components/Embed.tsx
```diff
@@ -81,6 +81,7 @@ const Embed: FC<UserProps> = ({ commonUser }) => {
 				poster={getFlowPlayerPoster(undefined, embedCode.content as ItemSchema)}
 				item={embedCode.content as ItemSchema}
 				canPlay={true}
+				initFlowPlayerWithoutAutoPlay={true}
```
bertyhell, line 84: Maybe a more clear name would be:
```suggestion
				placeholder={false}
```
Where placeholder would default to true, so we don't affect existing flowplayers already used by AVO

src/shared/components/FlowPlayerWrapper/FlowPlayerWrapper.tsx
```diff
@@ -363,16 +405,7 @@ const FlowPlayerWrapper: FC<
...
-							!isNil(end) &&
-							(start !== 0 || end !== toSeconds(item?.duration)) && (
-								<div className="c-cut-overlay">
-									<Icon name={IconName.scissors} />
-									{`${formatDurationHoursMinutesSeconds(
-										start
-									)} - ${formatDurationHoursMinutesSeconds(end)}`}
-								</div>
-							)}
+						{renderCutOverlay()}
```
bertyhell, line 408: nice!

src/shared/components/FlowPlayerWrapper/FlowPlayerWrapper.types.ts
```diff
@@ -32,6 +32,7 @@ export type FlowPlayerWrapperProps = {
 	controls?: boolean;
 	speed?: unknown | null;
 	trackPlayEvent: boolean;
+	initFlowPlayerWithoutAutoPlay?: boolean;
```
bertyhell, line 35:
```suggestion
	placeholder?: boolean;
```

bertyhell (changes requested):

PR #2121, opened 2025-06-23
AVO-3711: make sure the code injection is added after the bundle code…

reunefe:
… instead of before. Otherwise, component css will be overwritten with general css

### DIFF
--- a/vite.config.js
+++ b/vite.config.js
@@ -36,7 +36,14 @@ export default defineConfig(() => {
 		server: {
 			port: 8080,
 		},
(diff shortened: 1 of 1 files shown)


bertyhell:
is this something we should also include in the component libraries and the admin-core-ui library?

reunefe:
> is this something we should also include in the component libraries and the admin-core-ui library?

Maybe, although I don't think so. I suspect the CSS in those libraries are very specific already. The problem are the components that try to override those common classes and fail because the general css is loaded afterwards

PR #2123, opened 2025-06-26
Feature/avo 3710 replaced fragments

reunefe:


src/embed-code/components/EmbedContent.tsx
```diff
@@ -238,6 +236,18 @@ const EmbedContent: FC<EmbedProps & UserProps & EmbedFlowProps> = ({
 		);
 	};
 
+	const renderReplacementWarning = () => {
+		return (
+			item?.contentIsReplaced && (
+				<Alert type="danger" className="u-m-b-l">
+					{tHtml(
+						'Dit fragment werd uitzonderlijk vervangen door Het Archief voor Onderwijs. Het zou kunnen dat de tijdscodes of de beschrijving niet meer goed passen.'
```
bertyhell, line 244: is there no reason for replacement available in the database when an item is replaced? Shouldn't that reason be shown here?

bertyhell (approved):

src/embed-code/components/EmbedContent.tsx
reunefe: In none of the alerts where a fragment was replaced, was a reason mentioned to be shown. I just did the same thing as with collections

PR #2135, opened 2025-07-14
AVO-3738: publish specific error if klascement user id is not known

reunefe:


src/assignment/views/AssignmentEditMarcom.tsx
```diff
@@ -100,11 +100,16 @@ const AssignmentEditMarcom: FC<AssignmentEditMarcomProps & RouteComponentProps &
...
-					'assignment/views/assignment-edit-marcom___publiceren-naar-klascement-mislukt'
-				)
+			const avoError = tText(
+				'assignment/views/assignment-edit-marcom___publiceren-naar-klascement-mislukt'
 			);
+			const klascementError =
+				(err as any)?.innerException?.additionalInfo?.responseBody?.additionalInfo
+					?.klascementError?.uitzonderingen?.[0]?.diagnose ||
+				(err as any)?.innerException?.additionalInfo?.responseBody?.additionalInfo
+					?.klascementError;
```
bertyhell, line 110: since we use it in 2 locations it might be good to have a helper:
getReadableKlascementErrorMessage(err): string

bertyhell (approved):

PR #2150, opened 2025-08-04
AVO-3535: unify itsme and eID options and use callback for eID

reunefe:


bertyhell (approved):

src/authentication/helpers/redirects.ts
```diff
@@ -55,23 +55,6 @@ export function redirectToServerLoginPage(
...
-export function redirectToServerItsmeLogin(
-	location: RouteComponentProps['location'],
-	openInNewTab = false
-): void {
-	const returnToUrl = getRedirectUrl(location, openInNewTab);
-	const fullUrl = `${getEnv('PROXY_URL')}/auth/acmidm/login?${queryString.stringify({
-		returnToUrl,
-		authMech: 'itsme',
-		ltiJwtToken: EmbedCodeService.getJwtToken(),
-	})}`;
```
bertyhell, line 67: doesn't this auto select leerid now?
shouldn't this be a more general login screen instead of itsme or leerid specifc?

bertyhell (changes requested):

PR #2151, opened 2025-08-04
AVO-3485 and AVO-3486: check list of dropdown options and only render…

reunefe:
… a more options dropdown if we have more than 1 option

### DIFF
--- a/src/assignment/views/AssignmentDetail.tsx
+++ b/src/assignment/views/AssignmentDetail.tsx
@@ -543,7 +543,7 @@ const AssignmentDetail: FC<
 
 	// Render
 
-	const renderHeaderButtons = () => {
+	const renderAssignmentDropdownOptions = () => {
 		const ASSIGNMENT_DROPDOWN_ITEMS = [
 			...createDropdownMenuItem(
 				assignmentId,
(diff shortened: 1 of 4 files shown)


src/assignment/views/AssignmentDetail.tsx
```diff
@@ -573,6 +573,37 @@ const AssignmentDetail: FC<
 			),
 		];
 
+		if (ASSIGNMENT_DROPDOWN_ITEMS.length === 0) {
+			return;
```
bertyhell, line 577:
```suggestion
			return null;
```

src/collection/components/CollectionOrBundleEdit.tsx
```diff
@@ -1401,14 +1434,7 @@ const CollectionOrBundleEdit: FC<
...
 
 				{renderReorderButtons()}
-				<MoreOptionsDropdown
-					isOpen={isOptionsMenuOpen}
-					onOpen={() => setIsOptionsMenuOpen(true)}
-					onClose={() => setIsOptionsMenuOpen(false)}
-					label={getMoreOptionsLabel()}
-					menuItems={COLLECTION_DROPDOWN_ITEMS}
-					onOptionClicked={(optionId) => executeAction(optionId as CollectionMenuAction)}
-				/>
```
bertyhell, line 1411: Wouldn't it be handy if we had a MoreOptionsDropdownWrapper that had the zero, one and more than one rules build in?
or would that be overkill for the 3 usecases we have now?

reunefe: I was thinking the same thing with the same doubt 😅 But maybe adding it, would not be that bad for future cases

PR #2152, opened 2025-08-05
AVO-1444: Store columns in localstorage

reunefe:


src/admin/shared/components/FilterTable/FilterTable.utils.ts
```diff
@@ -11,3 +11,16 @@ export function cleanupObject(obj: any) {
...
+}
+
+export function setPreferredColumns(columnKey: string, value: string[]): void {
+	localStorage.setItem(getColumnKey(columnKey), JSON.stringify(value));
+}
+
+export function getPreferredColumns(columnKey: string): string[] {
+	const columns = localStorage.getItem(getColumnKey(columnKey));
+	return columns ? JSON.parse(columns) : [];
+}
```
bertyhell, line 26: you could consider to use the hook for this:
```
import { useLocalStorage } from '@uidotdev/usehooks';

const [isRemembered, setRemember] = useLocalStorage(rememberKey, false);
```

bertyhell (approved):

PR #2158, opened 2025-08-21
AVO-3774: show error of the proxy in case the external platform and t…

reunefe:
…he platform of the embed do not match


src/embed-code/embed-code-service.ts
```diff
@@ -44,17 +45,25 @@ export class EmbedCodeService {
 				credentials: 'include',
 			});
 
+			const responseJson = await response.json();
+
 			if (!response.ok) {
+				// THis means the external platform and the embed do not match
```
bertyhell, line 51:
```suggestion
				// This means the external platform and the embed do not match
```

bertyhell (approved):

PR #2159, opened 2025-08-26
AVO-3762: start moving assignment queries to proxy

reunefe:


bertyhell:
it's a bit suspicious that the generated database queries file is bigger than it was before this PR
src/shared/generated/graphql-db-types.ts

with all the queries you removed, you would expect it to be smaller? Any idea why that is?

bertyhell:
ah, ok, makes sense 👍

PR #2166, opened 2025-09-01
feat(AVO-3761): rework temp access to use profile id in db

bertyhell:
https://meemoo.atlassian.net/browse/AVO-3761

other PR's:
* https://github.com/viaacode/hasura-graphql/pull/142
* https://github.com/viaacode/avo2-proxy/pull/843

image


src/settings/components/Profile.tsx
```diff
@@ -461,9 +461,9 @@ const Profile: FC<
 			}
 
 			case 'temp_access': {
-				const tempAccess = profile?.user?.temp_access;
+				const tempAccess = profile?.temp_access;
 
-				return tempAccess?.current?.status === 1
+				return tempAccess?.has_currently_access?.status
```
reunefe, line 466: Just wondering why the check itself changed as well. Before it was checking whether or not the status is 1, but now it there is just a status?

reunefe (approved):

PR #2176, opened 2025-09-22
React router v7 en Jotai state management

bertyhell:
upgrade react router v5 to v7
replace redux with jotai
replace default imports with named exports where possible


bertyhell:
has already been merged into develop through another PR

